PR #1, opened 2024-07-31
data-sync: Added a base app with a build framework

RameshIyyar:
- Added a base build framework for the data synchronization application to enable in the redundant BMCs system to sync data.

- The synchronized data will be utilized by the respective BMC or host applications as needed.

- Refer design doc: https://gerrit.openbmc.org/c/openbmc/docs/+/71039


src/meson.build
```diff
@@ -0,0 +1,22 @@
...
+sdbusplus = dependency('sdbusplus')
+sdeventplus = dependency('sdeventplus')
+
+rbmc_data_sync_sources = [
+    'phosphor-rbmc-data-sync-mgr_main.cpp'
+  ]
+
+rbmc_data_sync_dependencies = [
+    phosphor_logging,
+    sdeventplus,
```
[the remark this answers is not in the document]
RameshIyyar: Done

OWNERS
```diff
@@ -0,0 +1,48 @@
...
+#
+# It is expected that these 4 sections will be listed in the order above and
+# data within them will be kept sorted.
+
+owners:
+- spinler@us.ibm.com
+- rameshi1@in.ibm.com
+
+reviewers:
+
```
spinler, line 39: May as well as Arya and Harsh here?

src/phosphor-rbmc-data-sync-mgr_main.cpp
```diff
@@ -0,0 +1,36 @@
...
+    try
+    {
+        using BMCDataSync =
+            sdbusplus::common::xyz::openbmc_project::data_sync::BMCData;
+
+        sdbusplus::async::context ctx;
+        sdbusplus::server::manager_t objManager{ctx,
+                                                BMCDataSync::namespace_path};
+
+        // clang-tidy currently mangles this into something unreadable
```
spinler, line 19: Since you don't have a .clang-tidy file to enable clang-tidy, I don't think you'd need to worry about the NOLINT stuff.

Though since you're starting from scratch, it may be a good time to add one, and just disable the checks that cause these complaints so the code doesn't have to worry about it.

(I was hoping they would alll be fixed in clang-19, but they weren't)

OWNERS
RameshIyyar: Sure, I will add them.

src/phosphor-rbmc-data-sync-mgr_main.cpp
RameshIyyar: Ah, I see. I misunderstood —I thought it was for Clang-Format and didn’t realize Clang-Tidy is different. Thanks for pointing that out! I’ll add clang tidy configuration.

src/meson.build
```diff
@@ -0,0 +1,25 @@
...
+  ]
+
+inc_dir = include_directories('.')
+libexecdir_installdir = join_paths(get_option('libexecdir'), 'phosphor-data-sync')
+executable('phosphor-rbmc-data-sync-mgr',
+    rbmc_data_sync_sources,
+    dependencies: rbmc_data_sync_dependencies,
+    include_directories: inc_dir,
+    install : true,
+    install_dir: libexecdir_installdir
```
aryakpadman, line 24: Why libexec is preferred as install_dir?

RameshIyyar: https://www.gnu.org/prep/standards/html_node/Directory-Variables.html

The data-sync application is not intended to be launched directly by users. Instead, it functions as a component of another application. Therefore, it is recommended to place it in the libexecdir directory.

harsh-agarwal1 (approved):

src/phosphor-rbmc-data-sync-mgr_main.cpp
```diff
@@ -0,0 +1,36 @@
...
+        // NOLINTNEXTLINE
+        ctx.spawn(
+            [](sdbusplus::async::context& ctx) -> sdbusplus::async::task<> {
+            ctx.request_name(BMCDataSync::interface);
+            co_return;
+        }(ctx));
+
+        ctx.run();
+    }
+    catch (std::exception& e)
```
spinler, line 29: this means you'll never get a core dump with a backtrace out of this app. I think that could be useful though to find the source of the exception?

src/phosphor-rbmc-data-sync-mgr_main.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+        sdbusplus::async::context ctx;
+        sdbusplus::server::manager_t objManager{ctx,
+                                                BMCDataSync::namespace_path};
+
+        // clang-tidy currently mangles this into something unreadable
+        // NOLINTNEXTLINE
+        ctx.spawn(
+            [](sdbusplus::async::context& ctx) -> sdbusplus::async::task<> {
+            ctx.request_name(BMCDataSync::interface);
```
spinler, line 23: I was looking at another review where the request_name isn't in a spawn: https://gerrit.openbmc.org/c/openbmc/dbus-sensors/+/75329/15/src/LeakDetection/LeakDetectionManager.cpp#104

not sure which way is better now.

src/meson.build
```diff
@@ -0,0 +1,25 @@
+# SPDX-License-Identifier: Apache-2.0
+
+phosphor_dbus_interfaces = dependency('phosphor-dbus-interfaces')
+phosphor_logging = dependency('phosphor-logging')
+sdbusplus = dependency('sdbusplus')
+
+rbmc_data_sync_sources = [
+    'phosphor-rbmc-data-sync-mgr_main.cpp'
```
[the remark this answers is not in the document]
RameshIyyar: Sure, I can rename it to `rbmc_data_sync_mgr.cpp`. I'm not sure how the `-` ended up being used, but I think we should stick to using `_` instead.

src/phosphor-rbmc-data-sync-mgr_main.cpp
RameshIyyar: I haven’t been too concerned about this now, but I’d prefer to revisit it later if we are persisting anything. Otherwise, we might encounter [this issue](https://github.com/ibm-openbmc/openpower-hw-isolation/blob/main/src/hardware_isolation_main.cpp#L44) that I discussed with Patrick a while ago.

In my opinion, there's no need to spawn here, but I’d prefer to revisit this if any issues arise related to the above.

src/phosphor-rbmc-data-sync-mgr_main.cpp
RameshIyyar: Typically, I prefer handling exceptions to ensure proper error management. However, I can add it to help catch potential human errors and improve error handling.

src/meson.build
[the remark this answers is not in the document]
aryakpadman: I feel keeping 'main' in file name would be better.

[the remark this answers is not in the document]
RameshIyyar: Sorry, Its typo...I could change into `rbmc_data_sync_main.cpp`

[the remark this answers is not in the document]
RameshIyyar: Done

src/phosphor-rbmc-data-sync-mgr_main.cpp
RameshIyyar: Done

aryakpadman (approved):

spinler (approved):

PR #2, opened 2024-10-25
data-sync: Adding data sync config files and validator

aryakpadman:
The changes include :-
    - Defines three JSON configuration files (common.json, open-power.json, and ibm.json) under config/data_sync_list.These files allow users to specify the files or directories to be synchronized between BMCs.

    - Creates a schema file (config/schema/schema.json) for the above JSON configuration files, adhering to the Draft 2020-12 schema specification.

    - Adds a Python script (scripts/validate_data_sync_list.py) to validate the JSON config files against the schema, along with another script (scripts/run-ci.sh) that triggers the validator during CI.

    - Adds a new Meson build option (data_sync_list) where users can provide the list of JSON configuration files without .json file extension.

    - Updates the Meson build system to install the specified JSON configuration files in the target directory and generate a config.h file containing all necessary build-time configurations.

Change-Id: I585880e27f87d9a0f9b557d4dfee2d18305e23ab

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -1,0 +1,30 @@
+# Data sync JSON configuration files
+
+The [JSON files](data_sync_list) specify the files and directories to be
+synchronized between the BMCs. Each configuration file must follow the schema
+defined in the [`config/schema/schema.json`](schema/schema.json).
+
+## Adding Files/Directories to Sync
+
+Users can either add files or directories for synchronization to one of the
+existing JSON files, if they fit in the categories described below, or create a
+new JSON file by following the schema in
+[`config/schema/schema.json`](schema/schema.json).
+
+Currently 3 JSON config files are defined under
+[`config/data_sync_list`](data_sync_list) as below.
+
+1. common.json - Contains files and directories of applications whic run across
+   all OpenBMC vendors.
+2. open-power.json - Contains files and directories of Open-Power applications.
+3. ibm.json - Contains files and directories specific to applications on IBM
+   systems.
+
+### Add a new config JSON file
+
+- If another vendor wants to add a new config JSON file into
+  [`config/data_sync_list`](data_sync_list) with their list of data to be
+  synced, define the schema compliant JSON file as `<vendor_name>.json`.
+- Update the `choices` field of `data_sync_list` option in
+  [meson.options](../meson.options) with the JSON file name without .json file
+  extension.
--- a/config/data_sync_list/common.json
+++ b/config/data_sync_list/common.json
@@ -1,0 +1,10 @@
+{
+    "Directories": [
+        {
+            "Path": "/var/lib/phosphor-state-manager/host{}-PersistData",
+            "Description": "Host State Persisted data",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Immediate"
+        }
+    ]
+}
--- a/config/data_sync_list/ibm.json
+++ b/config/data_sync_list/ibm.json
@@ -1,0 +1,10 @@
+{
+    "Files": [
+        {
+            "Path": "/var/lib/ibm/bmcweb/RootCert",
+            "Description": "Root certificate",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Immediate"
+        }
+    ]
+}
--- a/config/data_sync_list/open-power.json
+++ b/config/data_sync_list/open-power.json
@@ -1,0 +1,10 @@
+{
+    "Files": [
+        {
+            "Path": "/var/lib/phosphor-software-manager/hostfw/running/GUARD",
+            "Description": "GUARD",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Immediate"
+        }
+    ]
+}
--- a/config/schema/example.json
+++ b/config/schema/example.json
@@ -1,0 +1,40 @@
+{
+    "Files": [
+        {
+            "Path": "/file1/path/to/sync",
+            "Description": "Add details about the data and purpose of the synchronization",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Immediate",
+            "RetryAttempts": 1,
+            "RetryInterval": "PT10S"
+        },
+        {
+            "Path": "/file2/path/to/sync",
+            "Description": "Add details about the data and purpose of the synchronization",
+            "SyncDirection": "Bidirectional",
+            "SyncType": "Periodic",
+            "Periodicity": "PT10S",
+            "RetryAttempts": 2,
+            "RetryInterval": "PT10M"
+        }
+    ],
+    "Directories": [
+        {
+            "Path": "/directory1/path/to/sync",
+            "Description": "Add details about the data and purpose of the synchronization",
+            "SyncDirection": "Passive2Active",
+            "SyncType": "Periodic",
+            "Periodicity": "PT1H",
+            "RetryAttempts": 1,
+            "RetryInterval": "PT10M",
+            "ExcludeFilesList": ["/Path/of/files/must/be/ignored/for/sync"],
+            "IncludeFilesList": ["/Path/of/files/must/be/considered/for/sync"]
+        },
+        {
+            "Path": "/directory2/path/to/sync",
+            "Description": "Add details about the data and purpose of the synchronization",
+            "SyncDirection": "Bidirectional",
+            "SyncType": "Immediate"
+        }
+    ]
+}
--- a/config/schema/schema.json
+++ b/config/schema/schema.json
@@ -1,0 +1,187 @@
+{
+    "title": "Schema for the data sync JSON config files",
+    "$id": "https://github.com/ibm-openbmc/phosphor-data-sync/config/schema/schema.json",
+    "description": "The list of files and directories which need to be synced between BMCs",
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "type": "object",
+    "properties": {
+        "Files": {
+            "description": "List of the files to be synced and details of the mode of sync",
+            "type": "array",
+            "items": {
+                "$ref": "#/$defs/files"
+            },
+            "minItems": 1,
+            "uniqueItems": true
+        },
+        "Directories": {
+            "description": "List of the directories to be synced and details of the  mode of sync",
+            "type": "array",
+            "items": {
+                "$ref": "#/$defs/directories"
+            },
+            "minItems": 1,
+            "uniqueItems": true
+        }
+    },
+    "anyOf": [
+        {
+            "required": ["Files"]
+        },
+        {
+            "required": ["Directories"]
+        }
+    ],
+    "additionalProperties": false,
+
+    "$defs": {
+        "files": {
+            "type": "object",
+            "properties": {
+                "Path": {
+                    "$ref": "#/$defs/path"
+                },
+                "Description": {
+                    "$ref": "#/$defs/description"
+                },
+                "SyncDirection": {
+                    "$ref": "#/$defs/syncDirection"
+                },
+                "SyncType": {
+                    "$ref": "#/$defs/syncType"
+                },
+                "Periodicity": {
+                    "$ref": "#/$defs/periodicity"
+                },
+                "RetryAttempts": {
+                    "$ref": "#/$defs/retryAttempts"
+                },
+                "RetryInterval": {
+                    "$ref": "#/$defs/retryInterval"
+                }
+            },
+            "required": ["Path", "Description", "SyncDirection", "SyncType"],
+            "additionalProperties": false,
+            "allOf": [
+                { "$ref": "#/$defs/conditionForPeriodicity" },
+                { "$ref": "#/$defs/conditionForRetry" }
+            ]
+        },
+
+        "directories": {
+            "type": "object",
+            "properties": {
+                "Path": {
+                    "$ref": "#/$defs/path"
+                },
+                "Description": {
+                    "$ref": "#/$defs/description"
+                },
+                "SyncDirection": {
+                    "$ref": "#/$defs/syncDirection"
+                },
+                "SyncType": {
+                    "$ref": "#/$defs/syncType"
+                },
+                "Periodicity": {
+                    "$ref": "#/$defs/periodicity"
+                },
+                "RetryAttempts": {
+                    "$ref": "#/$defs/retryAttempts"
+                },
+                "RetryInterval": {
+                    "$ref": "#/$defs/retryInterval"
+                },
+                "ExcludeFilesList": {
+                    "$ref": "#/$defs/excludeFilesList"
+                },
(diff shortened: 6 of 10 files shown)


config/data_sync_list/common.json
aryakpadman, line 1: Would a README file be a good option to help users to understand the purpose of different JSON files?

aryakpadman:
Changing into draft as I felt some improvisations required after a self review. Will push the modified changes soon.

config/data_sync_list/common.json
spinler: Yes, I think that would be a good idea.

config/data_sync_list/common.json
```diff
@@ -0,0 +1,12 @@
+{
+    "RetryAttempts": 3,
```
spinler, line 2: Is there a way RetryAttempts and RetryInterval could have a good default?  I don't think we want everyone to have to decide these values, and instead just trust that the daemon will do the right thing.  especially because there's no reason for it to fail in the first place aside from something catastrophic so just a hiccup should be automatically handled.

config/schema/example.json
```diff
@@ -0,0 +1,36 @@
...
+    "Directories": [
+        {
+            "DirectoryName": "/directory1/path/to/sync",
+            "Description": "Add details about the data and purpose of the synchronization",
+            "SyncDirection": "Passive2Active",
+            "SyncType": "Periodic",
+            "Periodicity": "PT1H",
+            "RetryAttempts": 1,
+            "RetryInterval": "PT10M",
+            "ExcludeFilesList": ["/files/to/ignore/for/sync"],
```
spinler, line 32: can these by wildcards?

config/data_sync_list/common.json
```diff
@@ -0,0 +1,12 @@
...
+    "RetryInterval": "PT10M",
+    "Directories": [
+        {
+            "DirectoryName": "/var/lib/phosphor-state-manager/host{}-PersistData",
+            "Description": "Host State Persisted data",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Immediate"
```
spinler, line 9: Have you guys discussed the 'systemd reload after sync' option at all?

config/schema/example.json
aryakpadman: I also had the same thought. But in order to make the infrastructure ready ASAP, right now the plan is to support  only the path as value for 'ExcludeFilesList', 'IncludeFilesList', 'FileName' and 'DirectoryName' , and bring the support for regex later. Also no any use case noticed from the current list of files for syncing.

In the case of bringing the wildcard support, we should define the valid wildcard and its combination and also need to validate the same. By default JSON is not supporting the validation of wildcards. So need to explore on that part. 
The same points we need to consider while parsing also.

config/data_sync_list/common.json
aryakpadman: Had a re-thinking on this design. Also , defining retryAttempts and retryInterval on each JSON file will create conflicts. But defining a default value in daemon will loose the flexibility for different vendors for defining their default retry attempts if they wants. So thinking about an approach of defining a build option in meson and can be included the same values as macros in auto generated configuration file (config.h) .

meson.build
```diff
@@ -0,0 +1,39 @@
...
+        json_file,
+        install_dir : data_sync_config_dir
+    )
+
+endforeach
+
+#auto generate a config file with required build time configurations
+conf_data = configuration_data()
+
+conf_data.set_quoted('DATA_SYNC_CONFIG_DIR', get_option('prefix')+'/'+ data_sync_config_dir)
```
RameshIyyar, line 32: Please use `description` argument to add some details which will added as comment in the generated header file.

meson.build
```diff
@@ -0,0 +1,39 @@
...
+conf_data.set('COMMON_RETRY_ATTEMPTS', get_option('retry_attempts'))
+conf_data.set('COMMON_RETRY_INTERVAL', get_option('retry_interval'))
+
+configure_file(
```
RameshIyyar, line 36: I believe the new approach in Meson is to declare the dependency for the generated header file, ensuring that it is generated before building the source files that depend on it.

meson.options
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: Apache-2.0
+option(
```
RameshIyyar, line 2: Please add a comment to clarify that the choice must be the file name (without extension) from the config/data_sync_list directory.

meson.options
```diff
@@ -0,0 +1,22 @@
...
+    type : 'array',
+    choices : ['common','open-power','ibm'],
+    value : ['common'],
+    description : 'The set of files and directories to be synced within BMCs'
+)
+
+option(
+    'retry_attempts',
+    type : 'integer',
+    min : 0,
```
RameshIyyar, line 13: I think we should support a minimum value of 1, as 0 would mean no retries, correct?

Additionally, we should have a policy to disable retries in specific cases if needed. This can be tracked separately.

Please add description to explain this option usage.

meson.options
```diff
@@ -0,0 +1,22 @@
...
+    value : 3
+)
+
+option(
```
RameshIyyar, line 17: Please add description to explain this option usage.

meson.options
```diff
@@ -0,0 +1,22 @@
...
+    'retry_interval',
+    type : 'integer',
+    min : 0,
```
RameshIyyar, line 20: I assume 0 means, retry immediately?

scripts/run-ci.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
RameshIyyar, line 1: Add license details in all new files `# SPDX-License-Identifier: Apache-2.0`

scripts/validate_data_sync_list.py
```diff
@@ -0,0 +1,59 @@
...
+r"""
+The script which validates the list of JSON files using jsonschema module, which
+lists the files and directories to be synced between active and passive BMC.
+
+The script uses isoduration module for validating the time periods mentioned as
+per the ISO 8601 duration standard.
+"""
+
+
+def validate_schema(data_sync_list, schema_file):
```
RameshIyyar, line 16: Please add doxygen for this API.

scripts/validate_data_sync_list.py
```diff
@@ -0,0 +1,59 @@
...
+            with open(config_file) as config_file_handle:
+                config_file_json = json.load(config_file_handle)
+                jsonschema.validators.validate(
+                    config_file_json,
+                    schema_json,
+                    format_checker=jsonschema.Draft202012Validator.FORMAT_CHECKER,
+                )
+                print("Schema validation success for " + config_file)
+    except Exception as error:
+        sys.exit("Schema validation failed with error : " + str(error))
```
RameshIyyar, line 32: Please add file details in the trace.

config/README.md
```diff
@@ -0,0 +1,20 @@
+# Data sync JSON configuration files
+
+The JSON files `config/data\_sync\_list` specify the files and directories to be
```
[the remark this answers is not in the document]
aryakpadman: Tried. But again went something wrong. Will correct it asap.

config/README.md
```diff
@@ -0,0 +1,20 @@
...
+synchronized between BMCs. Each configuration file must follow the schema
+defined in `config/schema/schema.json`.
+
+## Adding Files/Directories to Sync
+
+Users can either add files or directories for synchronization to one of the
```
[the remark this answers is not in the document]
aryakpadman: I guess it conveys that also. User need to create new JSON files only if it is not fitting to any of the existing ones.

config/README.md
```diff
@@ -0,0 +1,20 @@
...
+JSON file following the schema in `config/schema/schema.json`.
+
+Currently 3 JSON config files are defined under `config/data\_sync\_list` as
+below.
+
+1. common.json - Contains files and directories of applications whic run across
+   all OpenBMC vendors.
+2. ibm.json - Contains files and directories specific to applications on IBM
+   systems.
+3. open-power.json - Contains files and directories of Open-Power applications.
```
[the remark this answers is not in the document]
aryakpadman: Modified.

meson.options
aryakpadman: Done

config/schema/example.json
```diff
@@ -0,0 +1,34 @@
+{
```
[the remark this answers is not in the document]
aryakpadman: Done

config/schema/schema.json
```diff
@@ -0,0 +1,219 @@
...
+            "required": ["Directories"]
+        }
+    ],
+    "additionalProperties": false,
+
+    "$defs": {
+        "files": {
+            "type": "object",
+            "properties": {
+                "FileName": {
```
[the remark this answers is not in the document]
aryakpadman: Modified the key to "Path" for both Files and Directories.

config/schema/schema.json
```diff
@@ -0,0 +1,219 @@
...
+            "description": "A short description about the file/directory to be synced",
+            "type": "string"
+        },
+        "syncDirection": {
+            "description": "The direction in which sync operation need to be performed.",
+            "enum": ["Active2Passive", "Passive2Active", "Bidirectional"]
+        },
+        "syncType": {
+            "description": "The type of sync to be performed",
+            "enum": ["Periodic", "Immediate"]
```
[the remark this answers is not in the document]
aryakpadman: Standard JSON is not supporting that. If it is too much confusing can either add in description . But as far as I know standard JSON also not supporting multi line strings also which will make description bulky. 

Last option might be adding a description in README file.

config/schema/schema.json
```diff
@@ -0,0 +1,219 @@
...
+                    "$ref": "#/$defs/syncDirection"
+                },
+                "SyncType": {
+                    "$ref": "#/$defs/syncType"
+                },
+                "Periodicity": {
+                    "$ref": "#/$defs/periodicity"
+                },
+                "RetryAttempts": {
+                    "description": "The number of retries for the specific file. This will override the common value",
```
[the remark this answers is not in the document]
aryakpadman: Done

config/schema/schema.json
```diff
@@ -0,0 +1,219 @@
...
+                "properties": {
+                    "Periodicity": {
+                        "$ref": "#/$defs/periodicity"
+                    }
+                },
+                "required": ["Periodicity"]
+            },
+            "else": {
+                "properties": {
+                    "Periodicity": { "not": {} }
```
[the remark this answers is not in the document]
aryakpadman: Yes . It will catch. The if..then..else clause is slightly modified for ease of readability.

config/schema/schema.json
```diff
@@ -0,0 +1,219 @@
...
+                    "RetryAttempts": { "type": "integer" }
+                },
+                "required": ["RetryAttempts"]
+            },
+            "then": {
+                "required": ["RetryInterval"]
+            },
+            "else": {
+                "properties": {
+                    "RetryInterval": { "not": {} }
```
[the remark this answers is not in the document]
aryakpadman: Yes

meson.build
```diff
@@ -0,0 +1,39 @@
...
+        install_dir : data_sync_config_dir
+    )
+
+endforeach
+
+#auto generate a config file with required build time configurations
+conf_data = configuration_data()
+
+conf_data.set_quoted('DATA_SYNC_CONFIG_DIR', get_option('prefix')+'/'+ data_sync_config_dir)
+conf_data.set('COMMON_RETRY_ATTEMPTS', get_option('retry_attempts'))
```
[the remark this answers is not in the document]
aryakpadman: Felt that it will help for better readability in source code, while dealing with file/directory specific retry.

meson.build
aryakpadman: Done

meson.options
aryakpadman: Done

scripts/validate_data_sync_list.py
aryakpadman: Done

scripts/validate_data_sync_list.py
aryakpadman: Done

meson.build
```diff
@@ -0,0 +1,45 @@
...
+
+
+foreach json_file_name : get_option('data_sync_list')
+
+    #check whether the json file given in the list exist and if so, install the same
+    json_file = files('config/data_sync_list/' + json_file_name + '.json')
+
+    install_data(
+        json_file,
+        install_dir : data_sync_config_dir
```
RameshIyyar, line 24: Should we also include the `prefix` during installation?

meson.build
RameshIyyar: I think it could be added into the respective commit.

aryakpadman (commented):
Addressed review comments

RameshIyyar (approved):

harsh-agarwal1 (approved):

config/data_sync_list/open-power.json
```diff
@@ -0,0 +1,16 @@
+{
+    "Directories": [
+        {
+            "Path": "/var/lib/phosphor-logging/errors",
```
spinler, line 4: There's going to need to be code changes in phosphor-logging before any syncing can be done, such as creating the filenames/log IDs based on the BMC position.  Otherwise, the errors will just overwrite the other one.

config/data_sync_list/ibm.json
```diff
@@ -0,0 +1,10 @@
+{
+    "Files": [
+        {
+            "Path": "/var/lib/ibm/bmcweb/RootCert",
+            "Description": "Root certificate",
+            "SyncDirection": "Active2Passive",
```
spinler, line 6: can bmcweb handle this being synced?

config/data_sync_list/open-power.json
aryakpadman: ok. I can remove this one as code changes is not there. Will add some other file details .Added some of the files/directories into each JSON files to enable the application framework.

config/data_sync_list/ibm.json
aryakpadman: Not just an example. Added the details as per the data we collected internally from every team.

spinler: Was wondering what bmcweb does when it has this file changed without it knowing.  I guess it will be a while before it becomes an issue so I won't worry about it now.

spinler (approved):

aryakpadman:
Rebased this commit with latest main branch.

config/data_sync_list/ibm.json
aryakpadman: Yeah I got you. For now we can proceed with this approach. Because my plan was just to make a basic infrastructure for data-sync now. Later we will add the details of all files and directories, and at that time will  ask the respective component owners to review the same again.

PR #4, opened 2024-12-20
DataSync: Design and implementation of config parser.

aryakpadman:


src/data_sync_config.cpp
```diff
@@ -0,0 +1,137 @@
...
+        lg2::error("Unsupported sync type [{SYNC_TYPE}]", "SYNC_TYPE",
+                   syncType);
+        return std::nullopt;
+    }
+}
+
+std::optional<std::chrono::seconds> DataSyncConfig::convertISODurationToSec(
+    const std::string& timeIntervalInISO)
+{
+    // TODO: Parse periodicity in ISO 8601 duration format using std::chrono
```
RameshIyyar, line 116: Could you clarify in comment, why you chose not to use `std::chrono`?

test/meson.build
```diff
@@ -0,0 +1,38 @@
+gtest_dep = dependency('gtest', main : true, required : false)
+if not gtest_dep.found()
+    gtest_proj = import('cmake').subproject('googletest', required: false)
```
RameshIyyar, line 3: The wrap file seems to be missing, so I think the subproject might not work.

test/meson.build
```diff
@@ -0,0 +1,38 @@
...
+        )
+    else
+        assert(
+            not get_option('tests').enabled(),
+            'Googletest is required if tests are enabled'
+            )
+    endif
+endif
+
+testSourceFiles = [
```
RameshIyyar, line 20: lowerCamelCase is not recommended by the meson, use snake case: Ref https://mesonbuild.com/Style-guide.html#naming-variable

the comment applicable to all places.

test/meson.build
```diff
@@ -0,0 +1,38 @@
...
+        'test_' + testFile.underscorify(),
+        executable(
+            'test-' + testFile.underscorify(),
+            testFile + '.cpp',
+            rbmc_data_sync_sources,
+            dependencies : [
+                gtest_dep,
+                rbmc_data_sync_dependencies,
+            ],
+            include_directories: include_directories('../src'),
```
RameshIyyar, line 35: I think we could use [inc_dir](https://github.com/ibm-openbmc/phosphor-data-sync/pull/4/commits/aee04830c645db9394d20f195059ca294373661a#diff-f28598af2e23aa5d2bc7c72e022ae2c56a33802eb970afffaeca1e40607f97feR23)? may be rename that variable into `src_inc_dir`?

src/meson.build
```diff
@@ -3,20 +3,27 @@
 phosphor_dbus_interfaces = dependency('phosphor-dbus-interfaces')
 phosphor_logging = dependency('phosphor-logging')
 sdbusplus = dependency('sdbusplus')
+nlohmann_json_dep = dependency('nlohmann_json')
 
 rbmc_data_sync_sources = [
-    'rbmc_data_sync_main.cpp'
+    files(
+        'data_sync_config.cpp',
+        'manager.cpp'
```
RameshIyyar, line 11: I noticed this file is missing from this commit but was added in a later commit. Please fix this. 

I hope you ran the local CI for each commit. (I know, working with multiple commit is pain :) )

src/manager.cpp
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: Apache-2.0
```
RameshIyyar, line 1: The Meson changes required to compile this file are missing. I left a comment about this in the previous commit.

src/manager.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <ranges>
+#include <string>
+#include <string_view>
+
+namespace data_sync
+{
+
+Manager::Manager(const fs::path& dataSyncCfgDir)
+{
+    parseConfiguration(dataSyncCfgDir);
```
RameshIyyar, line 20: I believe this could be converted into an async call, but it can be addressed separately based on the requirements.

src/manager.hpp
```diff
@@ -0,0 +1,56 @@
...
+     * @return NULL
+     *
+     * @note It will continue parsing all files even if one file fails to parse.
+     */
+    void parseConfiguration(const fs::path& dataSyncCfgDir);
+
+    /**
+     * @brief The list of data to synchronize.
+     */
+    std::vector<config::DataSyncConfig> _dataSyncConfiguration;
```
RameshIyyar, line 53: I think we could use a separate container to manage Active and Passive BMC files/directories independently, along with a separate handler. However, I haven’t fully explored the potential benefits yet. I’ll consider this in my subsequent commit (sync framework).

src/manager.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+void Manager::parseConfiguration(const fs::path& dataSyncCfgDir)
```
RameshIyyar, line 23: I think we could add a test case for this function, but I’m okay with skipping it since the API primarily uses standard library functions without much additional logic.

src/meson.build
```diff
@@ -1,8 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
-phosphor_dbus_interfaces = dependency('phosphor-dbus-interfaces')
-phosphor_logging = dependency('phosphor-logging')
-sdbusplus = dependency('sdbusplus')
+phosphor_dbus_interfaces_dep = dependency('phosphor-dbus-interfaces')
```
RameshIyyar, line 3: Thanks

RameshIyyar (changes requested):

aryakpadman (commented):
Addressing review comments.

test/meson.build
```diff
@@ -0,0 +1,21 @@
+gtest_dep = dependency('gtest', main : true, required : false)
```
RameshIyyar, line 1: The `required` parameter should be set to `false` to prevent the build from proceeding if the GTest library is not found in the build environment. It should only be set to `true` when a wrap file is explicitly defined.

test/data_sync_config_test.cpp
```diff
@@ -0,0 +1,135 @@
+// SPDX-License-Identifier: Apache-2.0
```
RameshIyyar, line 1: Just suggestion: It would be great if you added negative test cases by providing incorrect or unsupported values for sync type, direction, and duration interval.

aryakpadman: Done.

test/meson.build
aryakpadman: ok understood.. Addressed

RameshIyyar (approved):

src/manager.cpp
```diff
@@ -0,0 +1,67 @@
...
+            lg2::error("Failed to parse the configuration file : {CONFIG_FILE}",
+                       "CONFIG_FILE", configFile);
+        }
+    };
+
+    if (fs::exists(dataSyncCfgDir) && fs::is_directory(dataSyncCfgDir))
+    {
+        std::ranges::for_each(
+            fs::directory_iterator(dataSyncCfgDir),
+            [&parse](const auto& configFile) { parse(fs::path(configFile)); });
```
RameshIyyar, line 64: Sorry, I missed that..we could avoid lambda here, just use `parse` lambda itself? `fs::directory_iterator` will return [directory_entry](https://en.cppreference.com/w/cpp/filesystem/directory_entry) so we just use `path()` inside `parse` lambda.

src/manager.cpp
```diff
@@ -0,0 +1,67 @@
...
+                    std::back_inserter(this->_dataSyncConfiguration),
+                    [](const auto& element) {
+                    return config::DataSyncConfig(element);
+                });
+            }
+        }
+        catch (const std::exception& e)
+        {
+            // TODO Create error log
+            lg2::error("Failed to parse the configuration file : {CONFIG_FILE}",
```
RameshIyyar, line 55: Please include exception in the trace as well.

src/manager.cpp
aryakpadman: Addressed.

src/manager.cpp
aryakpadman: Addressed.

src/meson.build
```diff
@@ -1,22 +1,29 @@
...
```
spinler, line 3: I think it'd be good to add subprojects for all of these deps as well so the repo will always work doing standalone builds.  I use that all of the time, and pretty much all upstream repos have support for it (https://github.com/openbmc/openbmc/issues/3993)

src/data_sync_config.hpp
```diff
@@ -0,0 +1,196 @@
...
+struct Retry
+{
+    /**
+     * @brief The constructor
+     *
+     * @param[in] retryAttempts - The number of retries
+     * @param[in] retryIntervalInSec - The interval in which retry will
+     *                                 occur.
+     */
+    Retry(const uint8_t& retryAttempts,
```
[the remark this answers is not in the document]
aryakpadman: Const reference can avoid further copies as well can ensure that the data referenced is strictly read only, which I felt as more apt.

src/data_sync_config.hpp
```diff
@@ -0,0 +1,196 @@
...
+};
+
+/**
+ * @brief The structure contains data sync configuration specified
+ *        in the configuration file for each file or directory to be
+ *        synchronized between BMCs.
+ */
+struct DataSyncConfig
+{
+  public:
```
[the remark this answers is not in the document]
aryakpadman: I came to read that , compiler doesn’t create a default constructor when we define any type of constructor manually in the class. I verified the same behavior too, hence accidental invoking of DataSyncConfig() will throw error. If that is the case, do we need to explicitly delete the DataSyncConfig() ? 

Or do we have any other use case for deleting DataSyncConfig()?

Ref : https://www.geeksforgeeks.org/does-c-compiler-create-default-constructor-when-we-write-our-own/

src/manager.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * @class Manager
+ *
+ * @brief This class manages all configured data for synchronization
+ *        between BMCs.
+ */
+class Manager
+{
+  public:
```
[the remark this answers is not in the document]
aryakpadman: Same doubt as [this](https://github.com/ibm-openbmc/phosphor-data-sync/pull/4/commits/508031b30386a5dfc507c31255c7de28e2f81e6e#r1901113836)

src/meson.build
aryakpadman: Tried the subprojects approach and found that wrap file for the external dependency fails to clone as `libsystemd` dependency is not able to find which is required for sdbusplus. 

But added the wrap file the first line dependency packages in a separate commit. Will try to fix the issue ASAP. 
Note : Same error I observed in other repositories also. ( Eg : phosphor-logging)

test/meson.build
```diff
@@ -1,4 +1,21 @@
-gtest_dep = dependency('gtest', main : true, required : true)
+gtest_dep = dependency('gtest', main : true, required : false)
```
RameshIyyar, line 1: You need to use [disabler](https://mesonbuild.com/Reference-manual_functions.html#dependency_disabler) so it will work in the `if` check.

src/data_sync_config.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+
+Retry::Retry(const uint8_t& retryAttempts,
+             const std::chrono::seconds& retryIntervalInSec) :
+    _retryAttempts(retryAttempts), _retryIntervalInSec(retryIntervalInSec)
+{}
+
+DataSyncConfig::DataSyncConfig(const nlohmann::json& config) :
+    _path(config["Path"].get<std::string>()),
+    _syncDirection(convertSyncDirectionToEnum(config["SyncDirection"].get<std::string>())
```
RameshIyyar, line 21: I think, it crossing the max length per line...pls check the format.

src/data_sync_config.hpp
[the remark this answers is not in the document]
spinler: See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const

src/data_sync_config.hpp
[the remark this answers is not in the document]
spinler: yea you're right.  No changes needed.

src/meson.build
spinler: Some dependencies just need to be installed on your own system.

spinler: For systemd, you usually need to install systemd-dev or libsystemd-dev.

spinler (commented):
You'd also need a phosphor-logging wrap.

src/data_sync_config.cpp
aryakpadman: Addressed.

aryakpadman:
> You'd also need a phosphor-logging wrap.

src/data_sync_config.hpp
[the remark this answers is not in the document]
aryakpadman: ok understood. Addressed. 
Thanks for sharing the reference.

spinler (approved):

harsh-agarwal1 (approved):

RameshIyyar (approved):

geissonator:
add to approvelist

PR #5, opened 2025-01-04
Introduced a framework to retrieve external dependent data.

RameshIyyar:


test/manager_test.cpp
```diff
@@ -0,0 +1,79 @@
...
+                }
+            )"_json;
+
+        cfgFile << jsonData;
+    }
+
+    static void TearDownTestCase()
+    {
+        std::filesystem::remove_all(dataSyncCfgDir);
+        std::filesystem::remove(dataSyncCfgDir);
```
spinler, line 55: won't remove_all also remove the dir itself?

src/external_data_ifaces_impl.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+// NOLINTNEXTLINE
+sdbusplus::async::task<> ExternalDataIFacesImpl::fetchBMCRedundancyMgrProps()
+{
+    auto rbmcMgr = sdbusplus::async::proxy()
+                       .service(RBMC::interface)
+                       .path(RBMC::instance_path)
+                       .interface(RBMC::interface);
+
+    auto props =
```
spinler, line 22: The plan is to just let an exception thrown here crash the daemon? Not sure what else you could do, but just want to make sure you thought about it.

test/manager_test.cpp
RameshIyyar: Yeah, it’s not deleting. I’m a bit surprised too... it’s not like using `rm -rf`; it’s just meant to remove all its child files and subdirectories.

src/external_data_ifaces_impl.cpp
RameshIyyar: Yes, for now, I'm considering crashing the application, but I’d prefer a graceful exit if it fails to retrieve the external dependent data needed to initiate a sync operation. Additionally, I’d like to handle all errors separately using an error logging framework.

I noticed there's a set_error API in the sender/receiver proposal, but I didn’t see that incorporated into Patrick’s context design. So, I decided to handle this separately. I just realized I missed adding a TODO for this—I'll add it now.

test/manager_test.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include <filesystem>
+#include <fstream>
+
+#include <gtest/gtest.h>
+
+class ManagerTest : public ::testing::Test
+{
+  protected:
+    static void SetUpTestCase()
```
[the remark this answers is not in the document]
RameshIyyar: > can we use SetUpTestCase() also as setup API name ?

Yes,

> Because, I remeber gtest documentation prefers SetUpTestSuite() as the function name.
> Reference : http://google.github.io/googletest/advanced.html#sharing-resources-between-tests-in-the-same-test-suite

You're right, SetUpTestCase isn't very readable. I agree with your suggestion, so I'll use the name you recommended. Thanks for letting me know—it was added in Google Test v1.15.2, and since we're using the latest version, it works perfectly.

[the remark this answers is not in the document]
RameshIyyar: Done

src/external_data_ifaces_impl.cpp
RameshIyyar: Done

aryakpadman (approved):
Approving the PR

PR #10, opened 2025-01-17
Manager: Create periodic events for sync process

harsh-agarwal1:
- Creating sync events that will become active in specified time.  It's aimed at optimizing synchronization for files and directories that do not require immediate updates.
- By scheduling sync events at regular intervals, we can better handle frequently updated data where real-time synchronization
  isn't critical, such as audit logs and host console, will help to reduce load.
- Added utility functions to enable easy testing.

Tested:
- Verified the events are syncing data as required with 4 testcases:
  - PeriodicDataSyncTest: Write but stops before sync interval therefore no sync could take place.
  - PeriodicDataSyncDelayFileTest: Create a file after periodic sync is started, syncs new file after interval is passed.
  - PeriodicDataSyncMultiRWTest: Write multi time and check for sync taking place every time.
  - PeriodicDataSyncP2ATest: Checking passive to active sync.
- Verified behavior in simics environment.

Change-Id: I03aefb85f52d0c429bcb2441300fa1981cea5baa

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -131,8 +131,6 @@ void Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
 {
     using namespace std::string_literals;
     std::string syncCmd{"rsync --archive --compress"};
-
-    // Add source data path
     syncCmd.append(" "s + dataSyncCfg._path);
 
 #ifdef UNIT_TEST
@@ -160,11 +158,16 @@ sdbusplus::async::task<> Manager::monitorDataToSync(
     co_return;
 }
 
-// NOLINTNEXTLINE
-sdbusplus::async::task<> Manager::monitorTimerToSync(
-    [[maybe_unused]] const config::DataSyncConfig& dataSyncCfg)
+sdbusplus::async::task<>
+    // NOLINTNEXTLINE
+    Manager::monitorTimerToSync(const config::DataSyncConfig& dataSyncCfg)
 {
-    // TODO Create timer events to monitor data for sync
+    while (!_ctx.stop_requested())
+    {
+        co_await sdbusplus::async::sleep_for(
+            _ctx, dataSyncCfg._periodicityInSec.value());
+        syncData(dataSyncCfg);
+    }
     co_return;
 }
 
--- a/test/manager_test.cpp
+++ b/test/manager_test.cpp
@@ -1,66 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
-
-#include "manager.hpp"
-#include "mock_ext_data_ifaces.hpp"
-
-#include <sdbusplus/async/context.hpp>
-
-#include <filesystem>
-#include <fstream>
-
-#include <gtest/gtest.h>
-
-class ManagerTest : public ::testing::Test
-{
-  protected:
-    static void SetUpTestSuite()
-    {
-        char tmpdir[] = "/tmp/pdsCfgDirXXXXXX";
-        dataSyncCfgDir = mkdtemp(tmpdir);
-        std::filesystem::path dataSyncCfgFile{dataSyncCfgDir / "config.json"};
-        std::ofstream cfgFile(dataSyncCfgFile);
-
-        jsonData = R"(
-                {
-                    "Files": [
-                        {
-                            "Path": "/file/path/to/sync",
-                            "Description": "Parse test file",
-                            "SyncDirection": "Active2Passive",
-                            "SyncType": "Immediate"
-                        }
-                    ],
-                    "Directories": [
-                        {
-                            "Path": "/directory/path/to/sync",
-                            "Description": "Parse test directory",
-                            "SyncDirection": "Passive2Active",
-                            "SyncType": "Periodic",
-                            "Periodicity": "PT5M",
-                            "RetryAttempts": 1,
-                            "RetryInterval": "PT10M",
-                            "ExcludeFilesList": ["/directory/file/to/ignore"],
-                            "IncludeFilesList": ["/directory/file/to/consider"]
-                        }
-                    ]
-                }
-            )"_json;
-
-        cfgFile << jsonData;
-    }
-
-    static void TearDownTestSuite()
-    {
-        std::filesystem::remove_all(dataSyncCfgDir);
-        std::filesystem::remove(dataSyncCfgDir);
-    }
-
-    static std::filesystem::path dataSyncCfgDir;
-    static nlohmann::json jsonData;
-};
+#include "manager_test.hpp"
 
 std::filesystem::path ManagerTest::dataSyncCfgDir;
-nlohmann::json ManagerTest::jsonData;
+std::filesystem::path ManagerTest::tmpDataSyncDataDir;
+nlohmann::json ManagerTest::commonJsonData;
 
 TEST_F(ManagerTest, ParseDataSyncCfg)
 {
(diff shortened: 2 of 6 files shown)


src/manager.cpp
```diff
@@ -158,7 +158,13 @@ sdbusplus::async::task<> Manager::monitorDataToSync(
 sdbusplus::async::task<> Manager::monitorTimerToSync(
     [[maybe_unused]] const config::DataSyncConfig& dataSyncCfg)
 {
-    // TODO Create timer events to monitor data for sync
+    while (!_ctx.stop_requested())
+    {
+        co_await sdbusplus::async::sleep_for(
+            _ctx, std::chrono::seconds(dataSyncCfg._periodicityInSec.value()));
+        // Call rsync Wrapper
+        syncData(dataSyncCfg);
```
spinler, line 166: I wonder if you want to sync first, and then wait.  Only matters for the very first sync of course, so it happens right away.

harsh-agarwal1: We are doing a fullSync before starting these events. Therefore, felt it would be better to first wait then do the sync.

RameshIyyar (commented):
Some of the comment applies to all places.

src/manager.cpp
spinler: No, it should be fine.

test/manager_test.cpp
```diff
@@ -45,21 +44,69 @@ class ManagerTest : public ::testing::Test
...
+                                                    "commonconfig.json"};
+        std::ofstream cfgFile(dataSyncCommonCfgFile);
+        cfgFile << jsonData;
+    }
 
+    // Set up each individual test
+    void SetUp() override
+    {
+        dataSyncCfgFile = dataSyncCfgDir / "testconfig.json";
+        std::filesystem::remove(dataSyncCfgFile);
```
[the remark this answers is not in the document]
aryakpadman: +1.
Adding this logic in TearDown() makes more sense to me also as it acts as the destructor.

test/manager_test.cpp
```diff
@@ -17,9 +17,8 @@ class ManagerTest : public ::testing::Test
     {
         char tmpdir[] = "/tmp/pdsCfgDirXXXXXX";
         dataSyncCfgDir = mkdtemp(tmpdir);
-        std::filesystem::path dataSyncCfgFile{dataSyncCfgDir / "config.json"};
-        std::ofstream cfgFile(dataSyncCfgFile);
-
+        char tmpDataDir[] = "/tmp/pdsDataDirXXXXXX";
+        dataSyncDataDir = mkdtemp(tmpDataDir);
```
aryakpadman, line 21: Just a suggestion : Will `tmpDataSyncDataDir` make more sense?

harsh-agarwal1:
By default we are using all the json files present in the folder, hence some testcases are failing. Working on them.

test/manager_test.cpp
```diff
@@ -45,21 +44,69 @@ class ManagerTest : public ::testing::Test
...
+    {
+        dataSyncCfgFile = dataSyncCfgDir / "testconfig.json";
+        std::filesystem::remove(dataSyncCfgFile);
+    }
+
+    // Tear down each individual test
+    void TearDown() override
+    {
+        // Remove each item from the directory
+        for (const auto& entry :
```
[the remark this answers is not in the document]
harsh-agarwal1: This is deleting the directory as well.

test/manager_test.cpp
```diff
@@ -100,3 +147,196 @@ TEST_F(ManagerTest, ParseDataSyncCfg)
...
+
+    EXPECT_CALL(*mockExtDataIfaces, fetchSiblingBmcIP())
+        // NOLINTNEXTLINE
+        .WillRepeatedly([]() -> sdbusplus::async::task<> { co_return; });
+
+    EXPECT_CALL(*mockExtDataIfaces, fetchRbmcCredentials())
+        // NOLINTNEXTLINE
+        .WillRepeatedly([]() -> sdbusplus::async::task<> { co_return; });
+
+    nlohmann::json jsonData = {
```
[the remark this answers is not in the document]
harsh-agarwal1: CLang formatter doesn't allow for such changes.

aryakpadman:
**Regarding commit message :**

>such as less frequently updated data, will help to reduce load.

We are opting for periodic synchronization for data that are frequently updated, as continuous updates may increase the risk of inotify event flooding. So please update the commit message accordingly.  

**Suggestion:**
It would be more better to include test results from the production code testing rather than focusing on unit test (UT) perspective.

src/manager.cpp
```diff
@@ -7,6 +7,7 @@
```
RameshIyyar, line 7: Commit message:

> less frequently updated data

It’s not incorrect—we’re enabling this feature to handle frequently updated data, but there’s no need for immediate syncing. This includes data like host console traces and audit logs, which are primarily used for debugging. Since these traces update frequently, they can be synced periodically instead.

src/manager.cpp
```diff
@@ -131,7 +132,10 @@ void Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
 {
     using namespace std::string_literals;
     std::string syncCmd{"rsync --archive --compress"};
-
+    if (!std::filesystem::exists(dataSyncCfg._path))
```
RameshIyyar, line 135: I believe this check should not be added here, as it's a common API used for various sync types. Some of these may try to delete a file on a sibling BMC if it has been removed from the local BMC.

test/manager_test.cpp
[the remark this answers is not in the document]
RameshIyyar: Which API you used? remove/remove_all? Anyway its fine, trying to reduce the code if possible.

test/mock_ext_data_ifaces.hpp
```diff
@@ -16,6 +16,11 @@ class MockExternalDataIFaces : public ExternalDataIFaces
                 (override));
     MOCK_METHOD(sdbusplus::async::task<>, fetchSiblingBmcIP, (), (override));
     MOCK_METHOD(sdbusplus::async::task<>, fetchRbmcCredentials, (), (override));
+    void setBMCRole(BMCRole role)
```
RameshIyyar, line 19: > BMCRole role

const BMCRole& role

test/mock_ext_data_ifaces.hpp
```diff
@@ -16,6 +16,11 @@ class MockExternalDataIFaces : public ExternalDataIFaces
...
+    {
+        bmcRole(role);
```
RameshIyyar, line 21: Suggestion: LN 21 and 22 could be return as `return bmcRole(role);`

src/manager.cpp
```diff
@@ -130,9 +130,7 @@ sdbusplus::async::task<> Manager::startSyncEvents()
 void Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
 {
     using namespace std::string_literals;
-    std::string syncCmd{"rsync --archive --compress"};
-
-    // Add source data path
+    std::string syncCmd{"rsync --archive --compress --ignore-missing-args"};
```
RameshIyyar, line 133: > --ignore-missing-args

Why do we need this option? I couldn’t find any details in the commit message.

Also, if we are making changes to the rsync CLI, I would prefer them to be in a separate commit for better clarity.

This option suppresses errors when a file is missing, but I believe we should treat that as an error instead.

Are you encountering any errors during periodic sync when a file is missing? If so, we should handle it specifically in the periodic logic. One approach could be to generate an error and stop the periodic sync for that particular file. However, data sync is triggered automatically during BMC boot after the role is determined. At the same time, several other applications might start and configure their data in the data sync configuration to sync.

If data sync starts before those applications are starts, we should not treat this as an error. Instead, we should log a warning or debug message and let the process sleep according to the configured periodicity or we could use inotify framework to watch its creation and sync it then go for sleep as per configured interval for Periodic type sync.

harsh-agarwal1: If our source path is not present, the rsync command fails. To suppress it was using 'ignore-missing-args'. 
Will let the rsync to fail and get logged. 
Could be handled in future with --delete option and error handling based on [error code](https://stackoverflow.com/questions/20737204/comprehensive-list-of-rsync-error-codes).

RameshIyyar (approved):

aryakpadman (approved):

spinler (approved):

PR #12, opened 2025-02-05
Manager: Implement FullSync

manishtwari:
This contains:
  - FullSync functionality implementation.
  - GTest cases for validating FullSync behavior.
  

~~Note: Since the YAML changes are not part of the phosphor-dbus- interface this is expected to fail in the CI~~
~~Note: Seeing some undefined behaviour for Gtest on CI debugging that so I may push it for couple of time, apologies for that~~

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -127,7 +127,12 @@ sdbusplus::async::task<> Manager::startSyncEvents()
     co_return;
 }
 
-void Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
+// TODO: This isn't truly an async operation — Need to use popen/posix_spawn to
+// run the rsync command asynchronously but it will be handled as part of
+// concurrent sync changes.
+sdbusplus::async::task<bool>
+    // NOLINTNEXTLINE
+    Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
 {
     using namespace std::string_literals;
     std::string syncCmd{"rsync --archive --compress"};
@@ -147,7 +152,10 @@ void Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
         // TODO:
         // Retry and create error log and disable redundancy if retry is failed.
         lg2::error("Error syncing: {PATH}", "PATH", dataSyncCfg._path);
+
+        co_return false;
     }
+    co_return true;
 }
 
 // NOLINTNEXTLINE
@@ -166,7 +174,7 @@ sdbusplus::async::task<>
     {
         co_await sdbusplus::async::sleep_for(
             _ctx, dataSyncCfg._periodicityInSec.value());
-        syncData(dataSyncCfg);
+        co_await syncData(dataSyncCfg);
     }
     co_return;
 }
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -19,7 +19,7 @@ Manager::Manager(sdbusplus::async::context& ctx,
                  std::unique_ptr<ext_data::ExternalDataIFaces>&& extDataIfaces,
                  const fs::path& dataSyncCfgDir) :
     _ctx(ctx), _extDataIfaces(std::move(extDataIfaces)),
-    _dataSyncCfgDir(dataSyncCfgDir)
+    _dataSyncCfgDir(dataSyncCfgDir), _syncBMCDataIface(ctx, *this)
 {
     _ctx.spawn(init());
 }
@@ -29,6 +29,13 @@ sdbusplus::async::task<> Manager::init()
 {
     co_await sdbusplus::async::execution::when_all(
         parseConfiguration(), _extDataIfaces->startExtDataFetches());
+
+    // TODO: Explore the possibility of running FullSync and Background Sync
+    // concurrently
+    if (_extDataIfaces->bmcRedundancy())
+    {
+        co_await startFullSync();
+    }
 
     co_return co_await startSyncEvents();
 }
@@ -176,6 +183,62 @@ sdbusplus::async::task<>
             _ctx, dataSyncCfg._periodicityInSec.value());
         co_await syncData(dataSyncCfg);
     }
+    co_return;
+}
+
+// NOLINTNEXTLINE
+sdbusplus::async::task<void> Manager::startFullSync()
+{
+    _syncBMCDataIface.full_sync_status(FullSyncStatus::FullSyncInProgress);
+    lg2::info("Full Sync started");
+
+    auto fullSyncStartTime = std::chrono::steady_clock::now();
+
+    auto syncResults = std::vector<bool>();
+    size_t spawnedTasks = 0;
+
+    for (const auto& cfg : _dataSyncConfiguration)
+    {
+        if (isSyncEligible(cfg))
+        {
+            _ctx.spawn(
+                syncData(cfg) |
+                stdexec::then([&syncResults, &spawnedTasks](bool result) {
+                syncResults.push_back(result);
+                spawnedTasks--; // Decrement the number of spawned tasks
+            }));
+            spawnedTasks++;     // Increment the number of spawned tasks
+        }
+    }
+
+    while (spawnedTasks > 0)
+    {
+        co_await sdbusplus::async::sleep_for(_ctx,
+                                             std::chrono::milliseconds(50));
+    }
+
+    auto fullSyncEndTime = std::chrono::steady_clock::now();
+    auto FullsyncElapsedTime = std::chrono::duration_cast<std::chrono::seconds>(
+        fullSyncEndTime - fullSyncStartTime);
+
+    // If any sync operation fails, the FullSync will be considered failed;
+    // otherwise, it will be marked as completed.
+    if (std::ranges::all_of(syncResults,
+                            [](const auto& result) { return result; }))
+    {
+        _syncBMCDataIface.full_sync_status(FullSyncStatus::FullSyncCompleted);
+        lg2::info("Full Sync completed successfully");
+    }
+    else
+    {
+        _syncBMCDataIface.full_sync_status(FullSyncStatus::FullSyncFailed);
+        lg2::info("Full Sync failed");
+    }
+
+    // total duration/time diff of the Full Sync operation
+    lg2::info("Elapsed time for full sync: [{DURATION_SECONDS}] seconds",
+              "DURATION_SECONDS", FullsyncElapsedTime.count());
+
     co_return;
 }
 
--- a/src/manager.hpp
+++ b/src/manager.hpp
@@ -4,6 +4,7 @@
 
 #include "data_sync_config.hpp"
 #include "external_data_ifaces.hpp"
+#include "sync_bmc_data_ifaces.hpp"
 
 #include <filesystem>
 #include <ranges>
@@ -11,6 +12,9 @@
 
 namespace data_sync
 {
+
+using FullSyncStatus = sdbusplus::common::xyz::openbmc_project::control::
+    SyncBMCData::FullSyncStatus;
 
 namespace fs = std::filesystem;
 
@@ -52,6 +56,35 @@ class Manager
     bool containsDataSyncCfg(const config::DataSyncConfig& dataSyncCfg)
     {
         return std::ranges::contains(_dataSyncConfiguration, dataSyncCfg);
+    }
+
+    /**
+     * @brief Initiates a full synchronization between two BMCs.
+     *
+     *        - This method is responsible for initiating the  Full
+     *          synchronization process between two BMCs.
+     *        - The sync process is handled asynchronously.
+     *
+     */
+    sdbusplus::async::task<> startFullSync();
+
+    /**
+     * @brief Helper API that retrieves the sibling BMC IP and returns its
+     *        availability as a boolean.
+     *
+     * @return True if sibling BMC IP is empty; otherwise False.
+     */
+    bool isSiblingBmcNotAvailable()
+    {
+        return _extDataIfaces->siblingBmcIP().empty();
+    }
+
+    /**
+     * @brief Helper API fetches the full sync Dbus status-property.
+     */
+    FullSyncStatus getFullSyncStatus() const
+    {
+        return _syncBMCDataIface.full_sync_status();
     }
 
   private:
@@ -141,6 +174,11 @@ class Manager
      * @brief The list of data to synchronize.
      */
     std::vector<config::DataSyncConfig> _dataSyncConfiguration;
+
+    /**
+     * @brief SyncBMCData Server Interface object
+     */
+    dbus_ifaces::SyncBMCDataIface _syncBMCDataIface;
 };
 
 } // namespace data_sync
--- a/src/manager.hpp
+++ b/src/manager.hpp
@@ -85,8 +85,11 @@ class Manager
      *
      * @param[in] dataSyncCfg - The data sync config to sync
      *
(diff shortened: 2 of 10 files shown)


test/full_sync_test.cpp
```diff
@@ -0,0 +1,602 @@
...
+    ManagerTest::writeData(srcFile1, data1);
+
+    std::string data2{"Data written on the file2\n"};
+    ManagerTest::writeData(srcFile2, data2);
+
+    std::string data3{"Data written on the file3\n"};
+    ManagerTest::writeData(srcFile3, data3);
+
+    std::string data4{"Data written on the file4\n"};
+    // ManagerTest::writeData(srcFile4, data4);
```
RameshIyyar, line 412: Why is commented? or Add some comment like why are you doing this? I guess you are trying to avoid creating this file so that the fullsync will fail?

test/full_sync_test.cpp
```diff
@@ -0,0 +1,602 @@
...
+
+    data_sync::Manager manager{ctx, std::move(extDataIface),
+                               ManagerTest::dataSyncCfgDir};
+
+    auto waitingForFullSyncToFinish =
+        [&](sdbusplus::async::context& ctx) -> sdbusplus::async::task<void> {
+        auto status = manager.getFullsyncStatus();
+        while (status != FullSyncStatus::FullSyncInProgress)
+        {
+            std::cout << "Current sync status: " << static_cast<int>(status)
```
RameshIyyar, line 583: is it required?

test/full_sync_test.cpp
```diff
@@ -0,0 +1,602 @@
...
+            co_await sdbusplus::async::sleep_for(ctx,
+                                                 std::chrono::nanoseconds(200));
+            status = manager.getFullsyncStatus();
+        }
+
+        co_await sdbusplus::async::sleep_for(ctx,
+                                             std::chrono::microseconds(100));
+
+        EXPECT_EQ(status, FullSyncStatus::FullSyncInProgress)
+            << "FullSync status is Failed!";
```
RameshIyyar, line 594: > is Failed

is not FullSyncInProgress

test/full_sync_test.cpp
```diff
@@ -0,0 +1,602 @@
...
+        while (status != FullSyncStatus::FullSyncCompleted &&
+               status != FullSyncStatus::FullSyncFailed)
+        {
+            co_await sdbusplus::async::sleep_for(ctx,
+                                                 std::chrono::milliseconds(50));
+            status = manager.getFullsyncStatus();
+        }
+
+        EXPECT_EQ(status, FullSyncStatus::FullSyncFailed)
+            << "FullSync status is Failed!";
```
RameshIyyar, line 435: > is Failed

is not Failed

RameshIyyar (changes requested):

src/sync_bmc_data_ifaces.cpp
```diff
@@ -0,0 +1,59 @@
...
+        lg2::error(
+            "Error: Full Sync in progress. Operation cannot proceed at this time ");
+        throw sdbusplus::xyz::openbmc_project::Control::SyncBMCData::Error::
+            FullSyncInProgress();
+    }
+
+    co_return _ctx.spawn(_manager.startFullSync());
+}
+
+bool SyncBMCDataIfaces::set_property([[maybe_unused]] full_sync_status_t type,
```
spinler, line 48: Why is this even necessary?  You don't need to have functions for every single property, only for ones you want to change the behavior, which I don't think  see here.

src/manager.cpp
```diff
@@ -166,8 +178,70 @@ sdbusplus::async::task<>
...
+    return _dbusIfaces.full_sync_status();
+}
+
+// NOLINTNEXTLINE
+sdbusplus::async::task<void> Manager::startFullSync()
+{
+    // set the full sync status on the D-Bus property as FullSyncInProgress
+    _dbusIfaces.full_sync_status(FullSyncStatus::FullSyncInProgress);
+
+    std::mutex mtx;
```
spinler, line 206: I could be wrong here, but I thought mutexes weren't necessary as coroutines always suspend at known points?

abhilashraju: True, if this is going to be single-threaded, there is no need for it. Even if there are plans to schedule the sync in a thread pool scheduler, using a mutex is not a good idea. The purpose of the sender-receiver pattern is to avoid the explicit use of such primitives.

src/manager.cpp
```diff
@@ -166,8 +183,72 @@ sdbusplus::async::task<>
...
+{
+    _syncBMCDataIface.full_sync_status(FullSyncStatus::FullSyncInProgress);
+
+    auto fullSyncStartTime = std::chrono::utc_clock::now();
+
+    std::mutex mtx;
+    auto syncResults = std::vector<bool>();
+    int completedTasks = 0, spawnedTasks = 0;
+
+    for (const auto& cfg : _dataSyncConfiguration)
```
abhilashraju, line 207: Dont we have any when_all algorithm that accepts senders in a dynamic container such as vector?

src/manager.cpp
```diff
@@ -127,7 +137,12 @@ sdbusplus::async::task<> Manager::startSyncEvents()
     co_return;
 }
 
-void Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
+// TODO: This isn't truly an async operation — Need to use popen/posix_spawn to
```
abhilashraju, line 140: Correct me if I am wrong. Are we planing to spawn multiple process here? Is it going to be a fire and forget pattern? then how do we know if sync is successful or not? Is there any threshold for maximum number of parallel rsync process ?

RameshIyyar: WIP...Remember popen is child process and we could read from fd...

it would be something like below in high level

```
FILE *f = popen("./output", "r");
int fd = fileno(f);
fcntl(fd, F_SETFL, O_NONBLOCK);

fdio(fd)
```

src/manager.cpp
RameshIyyar: I think we could avoid mutex here while thinking more since we could decrease task counter in `then` of spawned sender...

it could be something like below..
```
for (const auto& cfg : _dataSyncConfiguration)
{
    if (isSyncEligible(cfg))
    {
        _ctx.spawn(this->syncData(cfg) | 
                   stdexec::then([&spawnedTasks](bool result) { --spawnedTasks; }));
         spawnedTasks++
    }
}

while (completedTasks != spawnedTasks)
{
        co_await sdbusplus::async::sleep_for(_ctx,
                                             std::chrono::milliseconds(50));
}
```

RameshIyyar: anything wrong here?

abhilashraju: it depends on what you do itn syncData. If it just popen and call the then, then you are not really finishing the work. you just spawned a process. You could have done it plain popen call directly in a loop. You have to just worry about how to monitor the returned fd using co_await, meaning replacing the while loop with some co_await .

src/manager.cpp
abhilashraju: Now you are spawning multiple processes in parallel instead of threads, which is more costly. It is better if you can use rsync as a library api call and use thread_pool scheduler to do it in parallel.

src/manager.cpp
```diff
@@ -15,11 +15,14 @@
 namespace data_sync
 {
 
+using FullSyncStatus = sdbusplus::common::xyz::openbmc_project::control::
```
RameshIyyar, line 18: Its already defined in the header file. Its not required.

src/manager.cpp
RameshIyyar, line 140: I think you pushed a separate commit for the same. Please fix it.

RameshIyyar: Yes, we understand it's costly, and just so you know, rsync doesn't have a library yet—it's planned for the future.

src/manager.cpp
```diff
@@ -166,8 +183,74 @@ sdbusplus::async::task<>
     {
         co_await sdbusplus::async::sleep_for(
             _ctx, dataSyncCfg._periodicityInSec.value());
-        syncData(dataSyncCfg);
+        co_await syncData(dataSyncCfg);
+    }
+    co_return;
+}
+
+FullSyncStatus Manager::getFullSyncStatus() const
```
RameshIyyar, line 191: We could move this definition into header file.

src/manager.cpp
RameshIyyar: @manishtwari Could you get a chance to check out this https://github.com/ibm-openbmc/phosphor-data-sync/pull/12/commits/c33edfe79b2a6f2ad2852be3f543da7c97441eb3#r1962074649 ?

test/manager_test.cpp
```diff
@@ -10,6 +10,40 @@ TEST_F(ManagerTest, ParseDataSyncCfg)
     using namespace std::literals;
     namespace ed = data_sync::ext_data;
 
+    commonJsonData = R"(
```
RameshIyyar, line 13: This is pushed into a separate commit, please fix it.

test/full_sync_test.cpp
```diff
@@ -0,0 +1,591 @@
...
+           {"DestinationPath",
+            ManagerTest::tmpDataSyncDataDir.string() + "/destFile3"},
+           {"Description", "FullSync from Passive to Active bmc"},
+           {"SyncDirection", "Passive2Active"},
+           {"SyncType", "Immediate"}},
+          {{"Path", ManagerTest::tmpDataSyncDataDir.string() + "/srcFile4"},
+           {"DestinationPath",
+            ManagerTest::tmpDataSyncDataDir.string() + "/destFile4"},
+           {"Description", "FullSync from Passive to Active bmc"},
+           {"SyncDirection", "Passive2Active"},
```
harsh-agarwal1, line 221: Let's make one file config as Active2Passive, and make sure it doesn't get synced

RameshIyyar (approved):

src/manager.cpp
```diff
@@ -179,4 +186,57 @@ sdbusplus::async::task<>
     co_return;
 }
 
+// NOLINTNEXTLINE
+sdbusplus::async::task<void> Manager::startFullSync()
+{
+    _syncBMCDataIface.full_sync_status(FullSyncStatus::FullSyncInProgress);
+
+    auto fullSyncStartTime = std::chrono::utc_clock::now();
```
spinler, line 194: steady_clock is usually better for keeping track of durations, because it is immune to wall clock changes.  I suspect utc_clock isn't.

src/manager.cpp
```diff
@@ -30,6 +30,13 @@ sdbusplus::async::task<> Manager::init()
     co_await sdbusplus::async::execution::when_all(
         parseConfiguration(), _extDataIfaces->startExtDataFetches());
 
+    // TODO: Explore the possibility of running FullSync and Background Sync
+    // concurrently
+    if (_extDataIfaces->bmcRedundancy())
+    {
+        co_await startFullSync();
```
spinler, line 37: This is just so if the service is restarted it can do a full sync to sync anything it missed?  because normally anytime this starts redundancy won't be enabled yet.

src/manager.cpp
```diff
@@ -179,4 +186,57 @@ sdbusplus::async::task<>
     co_return;
 }
 
+// NOLINTNEXTLINE
+sdbusplus::async::task<void> Manager::startFullSync()
+{
```
spinler, line 191: I think tracing the full sync start would be interesting data to have.  Then we could have to count backwards in the journal based  on how long it took to find it.

src/manager.hpp
```diff
@@ -54,6 +58,35 @@ class Manager
...
+     */
+    sdbusplus::async::task<> startFullSync();
+
+    /**
+     * @brief Helper API that retrieves the sibling BMC IP and returns its
+     *        availability as a boolean.
+     *
+     * @return True if sibling BMC IP is empty; otherwise False.
+     */
+    bool isSiblingBmcAvailable()
```
spinler, line 77: i think I'm confused on how an empty sibling IP means the sibling BMC is available, instead of the opposite.

src/sync_bmc_data_ifaces.cpp
```diff
@@ -0,0 +1,48 @@
...
+{
+    emit_added();
+}
+
+// NOLINTNEXTLINE
+sdbusplus::async::task<>
+    SyncBMCDataIfaces::method_call([[maybe_unused]] start_full_sync_t type,
+                                   [[maybe_unused]] sdbusplus::message_t& msg)
+{
+    if (_manager.isSiblingBmcAvailable())
```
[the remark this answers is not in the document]
spinler: this hasn't been resolved.

src/sync_bmc_data_ifaces.hpp
```diff
@@ -0,0 +1,68 @@
...
+                     data_sync::Manager& manager);
+
+    /**
+     * @brief Handles the FullSync method call for the SyncBmcData interface,
+     *
+     * @param[in] type Method type identifier.
+     * @param[in] msg D-Bus message containing the method call data.
+     */
+    sdbusplus::async::task<> method_call(start_full_sync_t type,
+                                         sdbusplus::message_t& msg);
```
spinler, line 53: where did you see that method_call() takes an sdbusplus::message_t param?

src/manager.cpp
```diff
@@ -166,9 +181,62 @@ sdbusplus::async::task<>
...
+
+    for (const auto& cfg : _dataSyncConfiguration)
+    {
+        if (isSyncEligible(cfg))
+        {
+            _ctx.spawn(
+                syncData(cfg) |
+                stdexec::then([&syncResults, &spawnedTasks](bool result) {
+                syncResults.push_back(result);
+                spawnedTasks--; // Decrement the number of spawned tasks
```
abhilashraju, line 207: Does this mean the sync is done or just started? If it is "just started," then line 220 gives incorrect data. It does not provide the sync completion time. Moreover, we should consider the retry attempts that impact the full sync time.

src/manager.cpp
```diff
@@ -166,9 +181,62 @@ sdbusplus::async::task<>
...
+            }));
+            spawnedTasks++;     // Increment the number of spawned tasks
+        }
+    }
+
+    while (spawnedTasks > 0)
+    {
+        co_await sdbusplus::async::sleep_for(_ctx,
```
abhilashraju, line 215: What does sleep_for do here? Does it make the thread sleep or just make the coroutine wait? Sleeping the thread makes the entire process sleep in a single-threaded application. Can't we use a timer for this?

manishtwari: It doesn’t actually make the thread or coroutine sleep. Instead, I’m using a timer
ref :
https://github.com/openbmc/sdbusplus/blob/e029082701db95ab6aa4cf89bbe9092445e9fd39/include/sdbusplus/async/timer.hpp#L118

src/manager.cpp
abhilashraju: I can see that right now syncData  invokes synchronous system call. Why don't you make simple while loop here? Or are you saying that ctx.spawn runs the syncdata in a different thread?

RameshIyyar: I think you know that spawn is single thread and we are spawning here to make track each sync going in parallel even its forked.

manishtwari: yes the idea here is that we spawn an independent task for each config, allowing syncData to run concurrently. right now we do have a synchronous system call, but as mentioned in the syncData TODO, we plan to use popen or posix_spawn to run the rsync command asynchronously in the future

spinler (approved):

PR #14, opened 2025-02-17
Manager: update syncData API

manishtwari:
- The syncData API needed to be updated because during a full sync we spawn a new config each time to handle it asynchronously, The existing implementation wasn’t designed to handle tasks running asynchronously, which is required for full sync operations

- To support this, the return type of syncData was changed to work with asynchronous execution, ensuring that tasks are handled concurrently

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -127,7 +127,12 @@ sdbusplus::async::task<> Manager::startSyncEvents()
     co_return;
 }
 
-void Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
+// TODO: This isn't truly an async operation — Need to use popen/posix_spawn to
+// run the rsync command asynchronously but it will be handled as part of
+// concurrent sync changes.
+sdbusplus::async::task<bool>
+    // NOLINTNEXTLINE
+    Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
 {
     using namespace std::string_literals;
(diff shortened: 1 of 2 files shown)


src/manager.cpp
```diff
@@ -147,7 +152,10 @@ void Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
         // TODO:
         // Retry and create error log and disable redundancy if retry is failed.
         lg2::error("Error syncing: {PATH}", "PATH", dataSyncCfg._path);
+
+        co_return false;
```
harsh-agarwal1, line 156: How about if we return the error code?

RameshIyyar: I think that's already planned as part of my rsync code. We need to manage all the defined error codes listed [here](https://download.samba.org/pub/rsync/rsync.1#EXIT_VALUES), and I think we can incorporate them into our error handling or retry logic.

If you prefer TODO then I think that could be added here.

RameshIyyar (approved):

aryakpadman (approved):

harsh-agarwal1 (approved):

RameshIyyar:
Please update your sign off as pointed in another PR so that we could merge #16 and #14

RameshIyyar (approved):

manishtwari:
This changes is not required, as the changes are already handled as part of the fullsync PR

PR #16, opened 2025-02-19
ManagerTest: Moved the commonJson

manishtwari:
- Moved the commonJson data from SetUpTestSuite to ParseDataSyncCfg because the file and directory paths in commonJson were not pointing to valid locations, causing errors when accessed, This commonJson is specific to ParseDataSyncCfg, so it was moved there.

 - Also, in the test case, we need to define the external JSON to simulate different scenarios where the commonJson is not required.

Change-Id: I0ae93189ed165ac0f077227cfc8760e6af5a1bfb

### DIFF
--- a/test/manager_test.cpp
+++ b/test/manager_test.cpp
@@ -10,6 +10,40 @@ TEST_F(ManagerTest, ParseDataSyncCfg)
     using namespace std::literals;
     namespace ed = data_sync::ext_data;
 
+    commonJsonData = R"(
+            {
+                "Files": [
+                    {
+                        "Path": "/file/path/to/sync",
+                        "Description": "Parse test file",
+                        "SyncDirection": "Active2Passive",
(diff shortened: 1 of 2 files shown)


spinler:
In the commit message:
```
Signed-off-by: manish-t-in ...
```
Can you fix that so it displays your real name?  It's a rule in OpenBMC.

And then also update in your other PRs if necessary.

RameshIyyar:
> In the commit message:
> 
> ```
> Signed-off-by: manish-t-in ...
> ```
> 
> Can you fix that so it displays your real name? It's a rule in OpenBMC.
> 
> And then also update in your other PRs if necessary.

+1

RameshIyyar (approved):

manishtwari:
This changes is not required, as the changes are already handled as part of the fullsync PR

PR #17, opened 2025-02-22
Implementation of immediate sync using inotify

aryakpadman:
Implementation of immediate sync using inotify

### DIFF
--- a/src/data_watcher.cpp
+++ b/src/data_watcher.cpp
@@ -13,6 +13,7 @@ namespace watch::inotify
 
 void DataWatcher::printWD(const std::map<int, fs::path>& _watchDescriptors)
 {
+    lg2::info("***Print current Watch descriptors***");
     for (const auto& [path, wd] : _watchDescriptors)
     {
         lg2::info("{PATH} --> {WD}", "PATH", path, "WD", wd);
@@ -26,13 +27,6 @@ DataWatcher::DataWatcher(sdbusplus::async::context& ctx, const int inotifyFlags,
     _fdioInstance(
         std::make_unique<sdbusplus::async::fdio>(ctx, _inotifyFileDescriptor()))
 {
-    if (!std::filesystem::exists(_dataPathToWatch))
-    {
-        lg2::info("Given path [{PATH}] doesn't exist to watch", "PATH",
-                  _dataPathToWatch);
-        // TODO : Handle not exists sceanrio by monitoring parent
-    }
-
     try
     {
         createWatchers(_dataPathToWatch, _eventMasksToWatch);
@@ -84,6 +78,7 @@ void DataWatcher::addToWatchList(const fs::path& currentPathToWatch,
     int wd = -1;
     wd = inotify_add_watch(_inotifyFileDescriptor(), currentPathToWatch.c_str(),
                            eventMasksToWatch);
+
     if (-1 == wd)
     {
         lg2::error(
@@ -130,7 +125,23 @@ void DataWatcher::createWatchers(const fs::path& currentPathToWatch,
     }
     else
     {
-        // TODO : If configured path not exist, monitor parent until it creates
+        lg2::info("Given path [{PATH}] doesn't exist to watch", "PATH",
+                  currentPathToWatch);
+
+        auto getExistingParentPath = [&](fs::path dataPath) {
+            while (!fs::exists(dataPath))
+            {
+                dataPath = dataPath.parent_path();
+            }
+            return dataPath;
+        };
+
+        if ((eventMasksToWatch & IN_CREATE) == 0)
+        {
+            eventMasksToWatch |= IN_CREATE;
+        }
+        addToWatchList(getExistingParentPath(currentPathToWatch),
+                       eventMasksToWatch);
     }
 }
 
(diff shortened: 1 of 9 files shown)


aryakpadman:
Facing `Timeout` Issue while executing test cases. Will fix it ASAP and will push next patch with UT also soon.

src/manager.cpp
```diff
@@ -130,11 +130,27 @@ sdbusplus::async::task<> Manager::startSyncEvents()
...
+        dataSyncCfg._destPath.value_or(dataSyncCfg._path)};
     std::string syncCmd{"rsync --archive --compress"};
-    syncCmd.append(" "s + dataSyncCfg._path);
+
+    if (isDelete)
+    {
+        std::string syncCmd{"rsync --delete --archive --compress"};
+        if (!std::filesystem::exists(srcPath))
+        {
+            srcPath = srcPath.parent_path();
```
aryakpadman, line 147: There can be cases where multiple sub directories got deleted. In that case instead of relying simply parent_path() , can find the existing parent path of that dataPath otherwise rsync will fail.

Will update this in next patch.

RameshIyyar:
add to approvelist

src/data_watcher.cpp
```diff
@@ -161,31 +166,133 @@ std::optional<std::vector<eventInfo>> DataWatcher::readEvents()
...
+            .string()
+            .starts_with(_dataPathToWatch.string()))
+    {
+        // Case 1 : The file configured in the JSON exists and is modified
+        // Case 2 : A file got created or modified inside a watching subdir
         return true;
     }
-    else if ((std::get<eventMask>(receivedEventInfo) &
-              (IN_CREATE | IN_ISDIR)) != 0)
+    else if (_watchDescriptors.at(std::get<0>(receivedEventInfo)) /
```
RameshIyyar: Use `using alias` for wd and basename. the same comment applicable to all places.

src/data_watcher.cpp
```diff
@@ -161,31 +166,133 @@ std::optional<std::vector<eventInfo>> DataWatcher::readEvents()
...
+    // all the file events are handled using IN_CLOSE_WRITE
+    if ((std::get<eventMask>(receivedEventInfo) & IN_ISDIR) != 0)
+    {
+        lg2::debug("Processing an IN_CREATE for {PATH}", "PATH",
+                   absCreatedPath);
+        if (absCreatedPath.string().starts_with(_dataPathToWatch.string()))
+        {
+            // The created dir is a child directory inside the configured data
+            // path add watch for the created child subdirectories.
+            fs::path subDirPath =
```
RameshIyyar, line 263: cannot we use `absCreatedPath`?

src/data_watcher.cpp
```diff
@@ -161,31 +166,133 @@ std::optional<std::vector<eventInfo>> DataWatcher::readEvents()
...
          */
-        lg2::debug("Processing an IN_CLOSE_WRITE for {PATH}", "PATH",
-                   _watchDescriptors.at(std::get<wd>(receivedEventInfo)) /
-                       std::get<basename>(receivedEventInfo));
+        return processCreate(receivedEventInfo);
+    }
+    return false;
+}
+
+bool DataWatcher::processCloseWrite(const eventInfo& receivedEventInfo)
```
RameshIyyar: Looks many places, you are trying to use `at` API to get something from map, why cannot get once and use it all places? The same comment is applicable to all commits (from the start).

src/data_watcher.cpp
```diff
@@ -161,31 +166,133 @@ std::optional<std::vector<eventInfo>> DataWatcher::readEvents()
...
+                        // Add configured event masks if created DIR is the
+                        // configured path.
+                        addToWatchList(entry, _eventMasksToWatch);
+                    }
+                    else
+                    {
+                        addToWatchList(entry, _eventMasksIfNotExists);
+                    }
+
+                    if (auto parent =
```
RameshIyyar, line 292: All these needed? cannot we simply remove the wd from the received event info?

src/data_watcher.cpp
```diff
@@ -161,31 +166,133 @@ std::optional<std::vector<eventInfo>> DataWatcher::readEvents()
...
+                modifyWatchIfExpected);
+            return true;
+        }
+    }
+    return false;
+}
+
+void DataWatcher::removeWatch(int wd)
+{
+    fs::path pathToRemove = _watchDescriptors.at(wd);
```
RameshIyyar, line 378: I believe we should minimize repeated accesses to the container. Instead, we should retrieve the data once and reuse it where needed.

For example, in this case, we're repeatedly accessing the map solely for debug tracing.

src/data_watcher.hpp
```diff
@@ -121,6 +121,13 @@ class DataWatcher
      */
     uint32_t _eventMasksToWatch;
 
+    /**
+     * @brief The group of events for which the data need to be watched on the
+     * parent if the configured path is not exisitng in the filesystem.
```
RameshIyyar, line 146: You could add details like why one more member with fixed event rather than using the given events to this class.

src/manager.cpp
```diff
@@ -173,7 +173,8 @@ sdbusplus::async::task<>
 {
     try
     {
-        uint32_t eventMasksToWatch = IN_CLOSE_WRITE;
+        uint32_t eventMasksToWatch = IN_CLOSE_WRITE | IN_DELETE |
```
RameshIyyar, line 176: IN_DELETE should be added for the directory, right?

src/manager.cpp
```diff
@@ -185,10 +186,12 @@ sdbusplus::async::task<>
 
         while (!_ctx.stop_requested())
         {
-            if (co_await dataWatcher.onDataChange())
-            {
-                co_await syncData(dataSyncCfg);
-            }
+            auto dataOperations = co_await dataWatcher.onDataChange();
```
RameshIyyar, line 189: it could be written as

```
if (auto dataOperations = co_await dataWatcher.onDataChange(); !dataOperations.empty())
{
std::ranges::for_each(dataOperations, [&dataSyncCfg]([[maybe_unused]] const auto& dataOp)
                -> sdbusplus::async::task<> { co_await syncData(dataSyncCfg); });
}
```

Why are we returning the operation but not using it? I suppose it would be useful to notify which specific file or directory has changed. This way, only the affected file or directory could be reloaded instead of reloading the entire configured directory. Additionally, it could help adjust the path for rsync to properly handle deletions, right?

src/data_watcher.cpp
```diff
@@ -219,15 +255,17 @@ bool DataWatcher::processCreate(const eventInfo& receivedEventInfo)
     {
         lg2::debug("Processing an IN_CREATE for {PATH}", "PATH",
                    absCreatedPath);
-        if (absCreatedPath.string().starts_with(_dataPathToWatch.string()))
+        if (absCreatedPath.string().starts_with(_dataPathToWatch.string()) &&
+            (_dataPathToWatch != absCreatedPath))
```
RameshIyyar, line 259: I think this check should be added in the respective commit, not here.

src/data_watcher.cpp
```diff
@@ -278,10 +316,61 @@ bool DataWatcher::processCreate(const eventInfo& receivedEventInfo)
...
+    // case 2 : A monitoring directory got deleted.
+    fs::path deletedPath =
+        _watchDescriptors.at(std::get<wd>(receivedEventInfo));
+    lg2::debug("Processing IN_DELETE_SELF for {PATH}", "PATH", deletedPath);
+
+    if (_watchDescriptors.size() == 1)
+    {
+        // Since no more watches will be there, add a watch on parent dir to
+        // notify future create events.
+        addToWatchList(getExistingParentPath(deletedPath),
```
RameshIyyar, line 338: I think `_watchDescriptors` will be more than one for the directory, right? If yes, not monitoring the parent directory of configured directory if its deleted?

aryakpadman (commented):
Addressed the review comments and fixed the CI issue of `TIMEOUT`. 

Also added a single test case which tests the immediate sync and is passing. Will enhance further.

src/data_watcher.hpp
```diff
@@ -0,0 +1,173 @@
...
+    FD(const FD&) = delete;
+    FD& operator=(const FD&) = delete;
+    FD(FD&&) = delete;
+    FD& operator=(FD&&) = delete;
+
+    /** @brief Saves File descriptor and uses it to do file operation
+     *
+     *  @param[in] fd - File descriptor
+     */
+    FD(int fd) : fd(fd) {}
```
spinler, line 72: nit:  The cppcheck tool that runs in CI (but doesn't cause fails) will complain if single argument constructors aren't explicit.  See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c46-by-default-declare-single-argument-constructors-explicit

src/data_watcher.hpp
```diff
@@ -0,0 +1,173 @@
...
+using EventMask = uint32_t;
+using EventInfo = std::tuple<BaseName, EventMask>;
+
+/** @class FD
+ *
+ *  @brief RAII wrapper for file descriptor.
+ */
+class FD
+{
+  private:
```
spinler, line 58: not a big deal, but better to be consistent on if public or private is first.  Usually public is.

src/data_watcher.hpp
```diff
@@ -0,0 +1,173 @@
...
+#include <sdbusplus/async.hpp>
+
+#include <filesystem>
+
+namespace data_sync
+{
+namespace watch::inotify
+{
+
+namespace fs = std::filesystem;
```
spinler, line 18: I don't think this is used in this file?  I wouldn't put it here just for other files to use, so either remove it or move it to the .cpp if you want to use it there.

src/data_watcher.cpp
```diff
@@ -103,7 +96,19 @@ void DataWatcher::createWatchers(const fs::path& pathToWatch)
     }
     else
     {
-        // TODO : If configured path not exist, monitor parent until it creates
+        lg2::debug("Given path [{PATH}] doesn't exist to watch", "PATH",
+                   pathToWatch);
+
+        auto getExistingParentPath = [&](fs::path dataPath) {
```
spinler: capturing everything by reference is frowned upon, better to just capture exactly what is needed, which I don't think is even anything?

src/data_watcher.cpp
```diff
@@ -156,31 +161,138 @@ std::optional<std::vector<EventInfo>> DataWatcher::readEvents()
...
+            (absCreatedPath != _dataPathToWatch))
+        {
+            // The created dir is a child directory inside the configured data
+            // path add watch for the created child subdirectories.
+            createWatchers(absCreatedPath);
+            return true;
+        }
+        else if (_dataPathToWatch.string().starts_with(absCreatedPath.string()))
+        {
+            // Was monitoring exisitng parent path of the configured data path
```
spinler, line 265: nit: spelling

src/data_watcher.cpp
```diff
@@ -278,10 +317,69 @@ bool DataWatcher::processCreate(const EventInfo& receivedEventInfo)
...
+        // emit for all sub directories which will remove its watches and
+        // finally will get IN_DELETE_SELF for the configured dir also which
+        // makes the size of _watchDescriptors 1.
+
+        addToWatchList(getExistingParentPath(deletedPath),
+                       _eventMasksIfNotExists);
+    }
+
+    // Remove the watch for the deleted path
+    auto wd = _watchDescriptors.begin();
```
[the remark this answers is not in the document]
aryakpadman: Agree.. Addressed

src/data_watcher.cpp
```diff
@@ -158,7 +164,22 @@ std::optional<std::vector<EventInfo>> DataWatcher::readEvents()
...
+    const std::vector<EventInfo>& receivedEventsInfo)
+{
+    std::ranges::for_each(receivedEventsInfo, [this](const auto& event) {
+        std::optional<DataOperation> dataOperation = processEvent(event);
+        if (dataOperation.has_value())
+        {
+            _dataOperations.insert_or_assign(dataOperation.value().first,
+                                             dataOperation.value().second);
+        }
+        dataOperation.reset();
```
RameshIyyar, line 177: nit: Its not required but nothing wrong.

RameshIyyar (approved):

src/data_watcher.cpp
aryakpadman: understood..removed

aryakpadman:
While testing observed some junk values in the buffer which stores the inotify event data . Hence pushed the change to do a memset on the buffer with `\0` also.

spinler (approved):

PR #18, opened 2025-02-27
Manager: Implement Disable Sync property

harsh-agarwal1:
- This commit provides implementation of disable sync DBus property.
- If disable is set to true, we restrict starting FullSync and stop all future background sync.
- If this property is set to false (i.e., sync enabled), we trigger sync events (immediate and periodic).
- Tested on simics.

~~This PR cannot get merged, before [PR168](https://github.com/ibm-openbmc/phosphor-dbus-interfaces/pull/168) for dbus-interfaces gets merged.~~

Change-Id: Ic7a6a63c1da74102405801d9e265c465f8a29af6

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -30,6 +30,13 @@ sdbusplus::async::task<> Manager::init()
     co_await sdbusplus::async::execution::when_all(
         parseConfiguration(), _extDataIfaces->startExtDataFetches());
 
+    if (_syncBMCDataIface.disable_sync())
+    {
+        lg2::info(
+            "Sync is Disabled, data sync cannot be performed to the sibling BMC.");
+        co_return;
+    }
+
     // TODO: Explore the possibility of running FullSync and Background Sync
     // concurrently
     if (_extDataIfaces->bmcRedundancy())
@@ -177,13 +184,35 @@ sdbusplus::async::task<>
     // NOLINTNEXTLINE
     Manager::monitorTimerToSync(const config::DataSyncConfig& dataSyncCfg)
 {
-    while (!_ctx.stop_requested())
+    while (!_ctx.stop_requested() && !_syncBMCDataIface.disable_sync())
     {
         co_await sdbusplus::async::sleep_for(
             _ctx, dataSyncCfg._periodicityInSec.value());
+        // Below is temporary check to avoid sync when disable sync is set to
+        // true.
+        // TODO: add receiver logic to stop sync events when disable sync is set
+        // to true.
+        if (_syncBMCDataIface.disable_sync())
+        {
+            break;
+        }
         co_await syncData(dataSyncCfg);
     }
     co_return;
+}
+
+void Manager::disableSyncPropChanged(bool disableSync)
+{
+    if (disableSync)
+    {
+        // TODO: Disable all sync events using Sender Receiver.
+        lg2::info("Sync is Disabled, Stopping events");
+    }
+    else
+    {
+        lg2::info("Sync is Enabled, Starting events");
+        _ctx.spawn(startSyncEvents());
+    }
 }
 
 // NOLINTNEXTLINE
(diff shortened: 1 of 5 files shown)


src/manager.cpp
```diff
@@ -30,6 +30,13 @@ sdbusplus::async::task<> Manager::init()
     co_await sdbusplus::async::execution::when_all(
         parseConfiguration(), _extDataIfaces->startExtDataFetches());
 
+    if (_syncBMCDataIface.disable_sync())
```
RameshIyyar, line 33: We need to persist this property, but it could be managed as part of the data sync DBus properties persistence work.

RameshIyyar (approved):
LGTM...

RameshIyyar:
jenkins run tests please

test/manager_test.cpp
```diff
@@ -78,3 +78,76 @@ TEST_F(ManagerTest, ParseDataSyncCfg)
...
+                               ManagerTest::dataSyncCfgDir};
+    manager.setDisableSyncStatus(true); // disabled the sync events
+
+    EXPECT_NE(ManagerTest::readData(destFile), data)
+        << "The data should not match because the manager is spawned and"
+        << " is waiting for the periodic interval to initiate the sync.";
+
+    ctx.spawn(sdbusplus::async::sleep_for(ctx, 1.1s) |
+              sdbusplus::async::execution::then(
+                  [&srcFile, &destFile, &data, &manager]() {
```
RameshIyyar, line 139: > &srcFile

Unused variable, CI is failed. I hope you are running the local CI.

RameshIyyar (changes requested):

src/manager.cpp
```diff
@@ -177,15 +184,37 @@ sdbusplus::async::task<>
...
+void Manager::disableSyncPropChanged(bool disableSync)
+{
+    if (disableSync)
+    {
+        // TODO: Disable all sync events using Sender Receiver.
+        lg2::info("Sync is Disabled, Stopping events");
+    }
+    else
+    {
+        lg2::info("Sync is Enabled, Starting events");
```
spinler, line 213: The sequence of this being used would probably be like:

1. Disable sync due to something bad happening
2. Re-enable sync due to some recovery
3. Do full sync to get what was missed

This code here is starting background syncs at 2., so is it OK to run a full sync and background sync in parallel?

harsh-agarwal1: We had the same doubt; to be sure, we tried this scenario on Simics. 
We enabled sync and immediately started full sync; it was working fine.

RameshIyyar: > This code here is starting background syncs at 2., so is it OK to run a full sync and background sync in parallel?

We added a [TODO](https://github.com/ibm-openbmc/phosphor-data-sync/blob/main/src/manager.cpp#L33) to explore running in parallel to evaluate the potential benefits.

However, I want to confirm my understanding of your question. Are you asking whether data sync can perform a full sync on its own after starting (or in parallel with) the background sync?

spinler: If RBMC manager were to ever disable sync for some reason, it may need to re-enable it again later. (this starts background sync).  And then to sync what was missed while it was off, it would have to do a full sync.  So that means we'd be requesting a full sync while background sync was running.

I was just asking if that was OK.

RameshIyyar: I wouldn't prefer that, but I'm curious to see what happens if it runs in parallel.

Also, I'm considering how we could make it fully synchronous first before starting the event. Allowing full sync even when sync is disabled doesn't seem logical to me. Should we instead define "disable sync" as only affecting background sync? That way, the RBMC manager could perform a full sync first and then enable sync if the full sync succeeds (or even if it fails?).

spinler: We would want to avoid any sort of delay in between the full sync completing and starting the background sync, otherwise changes could be missed.

harsh-agarwal1: Once immediate sync code is merged, we can try to mimic the parallel sync together for all type + full sync and observe what happening.

RameshIyyar: I believe we should include this in the upcoming performance evaluation, so we can revisit it if any issues arise.

Matt, I hope you are fine with it.

RameshIyyar (approved):

spinler (approved):

PR #19, opened 2025-03-07
Manager: Update disable sync with Immediate sync code

harsh-agarwal1:
- Modified disable sync property logic to align with the latest immediate sync code.
- Moved PeriodicDisablePropertyTest from manager_test to periodic_sync_test to avoid conflicts with common JSON file config.
- Added 'testDataChangeDisableProperty' test case.
- Tested on simics.

Change-Id: Id2c590a51c926d6d11eccf5c4aacc51298788fb8


src/manager.cpp
```diff
@@ -190,11 +190,19 @@ sdbusplus::async::task<>
         watch::inotify::DataWatcher dataWatcher(
```
RameshIyyar, line 190: Commit message:

Header: `Manager: Update disable sync with latest code` 

Please avoid this type of commit message header. We've discussed this pattern previously as well. Please describe what are you doing in the commit message header and why should be in commit message body as per guideline (https://github.com/openbmc/docs/blob/master/CONTRIBUTING.md#submitting-changes). 

> - Moved PeriodicDisablePropertyTest from manager_test to
  periodic_sync_test to avoid conflicts with common JSON file config.

It should be in separate commit and why are we moving need to capture.

> - Added 'testDataChangeDisableProperty' test case.
Not required to capture UT cases details.

> - Tested on simics.

Please elaborate like how do you test in the simics.

test/immediate_sync_test.cpp
```diff
@@ -88,3 +88,90 @@ TEST_F(ManagerTest, testDataChangeInFile)
 
     ctx.run();
 }
+
+TEST_F(ManagerTest, testDataChangeDisableProperty)
```
RameshIyyar, line 92: > testDataChangeDisableProperty

I think its not clear. looks you are trying to test data change when the sync is disabled/

How about `testDataChangeWhenSyncIsDisabled`?

test/immediate_sync_test.cpp
```diff
@@ -88,3 +88,90 @@ TEST_F(ManagerTest, testDataChangeInFile)
...
+    EXPECT_CALL(*mockExtDataIfaces, fetchRbmcCredentials())
+        // NOLINTNEXTLINE
+        .WillRepeatedly([]() -> sdbusplus::async::task<> { co_return; });
+
+    nlohmann::json jsonData = {
+        {"Files",
+         {{{"Path", ManagerTest::tmpDataSyncDataDir.string() + "/srcFile2"},
+           {"DestinationPath",
+            ManagerTest::tmpDataSyncDataDir.string() + "/destFile2"},
+           {"Description", "File to test immediate sync upon data write"},
```
RameshIyyar, line 123: Please update as per test case scope.

test/immediate_sync_test.cpp
```diff
@@ -88,3 +88,90 @@ TEST_F(ManagerTest, testDataChangeInFile)
...
+    EXPECT_NE(ManagerTest::readData(destPath), data)
+        << "The data should not match because the manager is spawned and"
+        << " sync is disabled.";
+
+    std::string dataToWrite{"Data is modified"};
+    std::string dataToStopEvent{"Close spawned inotify event."};
+
+    // write data to src path to create inotify event so that the spawned
+    // process is closed.
+    ManagerTest::writeData(srcPath, dataToStopEvent);
```
RameshIyyar, line 154: is it guarantee that background sync will be ready to catch this? I think we added some delay in the previous commits to guarantee that.

test/immediate_sync_test.cpp
```diff
@@ -88,3 +88,90 @@ TEST_F(ManagerTest, testDataChangeInFile)
...
+    // Write data after 2s so that the background sync events will be ready
+    // to catch.
+    ctx.spawn(
+        sdbusplus::async::sleep_for(ctx, 2s) |
+        sdbusplus::async::execution::then([&ctx, &srcPath, &dataToWrite]() {
+        ManagerTest::writeData(srcPath, dataToWrite);
+        ctx.request_stop();
+    }));
+    ctx.run();
+    EXPECT_EQ(ManagerTest::readData(destPath), dataToWrite)
```
RameshIyyar, line 174: Please check other test cases. We are watching dest path before checking the expectation.

test/immediate_sync_test.cpp
```diff
@@ -88,3 +88,90 @@ TEST_F(ManagerTest, testDataChangeInFile)
...
+        EXPECT_NE(ManagerTest::readData(destPath), dataToStopEvent)
+            << "The data should not match as sync is disabled even though"
+            << " sync should take place at every data change.";
+        manager.setDisableSyncStatus(false); // Trigger the sync events
+    }));
+
+    // Write data after 2s so that the background sync events will be ready
+    // to catch.
+    ctx.spawn(
+        sdbusplus::async::sleep_for(ctx, 2s) |
```
RameshIyyar, line 168: Could you confirm if 2 seconds and above 1 second are necessary? I’d like to reduce the sleep interval, if possible, to decrease the overall unit test execution time.

test/manager_test.cpp
```diff
@@ -78,75 +78,3 @@ TEST_F(ManagerTest, ParseDataSyncCfg)
     EXPECT_TRUE(manager.containsDataSyncCfg(data_sync::config::DataSyncConfig(
         ManagerTest::commonJsonData["Files"][0], false)));
 }
-
-TEST_F(ManagerTest, PeriodicDisablePropertyTest)
```
RameshIyyar, line 82: Why its removed?

test/immediate_sync_test.cpp
harsh-agarwal1: This test case has different requirement, as we are checking its value against `dataToStopEvent` and `dataToWrite`, once expecting not equal, 2nd time expecting equal. Therefore, cannot have single check for `destPath` whenever data is changed.

spinler (approved):

test/immediate_sync_test.cpp
RameshIyyar: I don't see anything major here, but we should still monitor dest for your exception check.

RameshIyyar (approved):

PR #20, opened 2025-03-12
Rsync: Updating flags in rsync CLI 

riya461:
- Fixing the full sync test case to sync in the expected directory 
- Updated test cases handling `--delete` and `--delete-missing-args` flags 
- Updated path comparison in `DataWatcher` from operator overloading to file system `equivalent` API
- Adding `/` to the end of directory path for expected syncing

### DIFF
--- a/src/data_watcher.cpp
+++ b/src/data_watcher.cpp
@@ -240,14 +240,14 @@ std::optional<DataOperation>
 std::optional<DataOperation>
     DataWatcher::processCreate(const EventInfo& receivedEventInfo)
 {
-    fs::path absCreatedPath =
-        _watchDescriptors.at(std::get<WD>(receivedEventInfo)) /
-        std::get<BaseName>(receivedEventInfo);
-
     // Process IN_CREATE only for DIR and skip for files as
     // all the file events are handled using IN_CLOSE_WRITE
     if ((std::get<EventMask>(receivedEventInfo) & IN_ISDIR) != 0)
     {
+        fs::path absCreatedPath =
+            _watchDescriptors.at(std::get<WD>(receivedEventInfo)) /
+            std::get<BaseName>(receivedEventInfo) / "";
+
         lg2::debug("Processing an IN_CREATE for {PATH}", "PATH",
                    absCreatedPath);
         if (absCreatedPath.string().starts_with(_dataPathToWatch.string()) &&
--- a/test/immediate_sync_test.cpp
+++ b/test/immediate_sync_test.cpp
@@ -343,3 +343,74 @@ TEST_F(ManagerTest, testDataChangeWhenSyncIsDisabled)
         << "The data should match with the data as the src was modified"
         << " and sync should take place at every modification.";
 }
+
+TEST_F(ManagerTest, testDataCreateInSubDir)
+{
+    using namespace std::literals;
+    namespace extData = data_sync::ext_data;
+
+    std::unique_ptr<extData::ExternalDataIFaces> extDataIface =
+        std::make_unique<extData::MockExternalDataIFaces>();
+
+    extData::MockExternalDataIFaces* mockExtDataIfaces =
+        dynamic_cast<extData::MockExternalDataIFaces*>(extDataIface.get());
+
+    ON_CALL(*mockExtDataIfaces, fetchBMCRedundancyMgrProps())
+        // NOLINTNEXTLINE
+        .WillByDefault([&mockExtDataIfaces]() -> sdbusplus::async::task<> {
+        mockExtDataIfaces->setBMCRole(extData::BMCRole::Active);
+        co_return;
+    });
+
+    EXPECT_CALL(*mockExtDataIfaces, fetchSiblingBmcIP())
+        // NOLINTNEXTLINE
+        .WillRepeatedly([]() -> sdbusplus::async::task<> { co_return; });
+
+    EXPECT_CALL(*mockExtDataIfaces, fetchRbmcCredentials())
+        // NOLINTNEXTLINE
+        .WillRepeatedly([]() -> sdbusplus::async::task<> { co_return; });
+
+    nlohmann::json jsonData = {
+        {"Directories",
+         {{{"Path", ManagerTest::tmpDataSyncDataDir.string() + "/srcDir/"},
+           {"DestinationPath",
+            ManagerTest::tmpDataSyncDataDir.string() + "/destDir/"},
+           {"Description",
+            "File to test immediate sync on non existent dest path"},
+           {"SyncDirection", "Active2Passive"},
+           {"SyncType", "Immediate"}}}}};
+
+    std::string srcDir{jsonData["Directories"][0]["Path"]};
+    std::string destDir{jsonData["Directories"][0]["DestinationPath"]};
+
+    // Create directories in source and destination
+    std::filesystem::create_directory(srcDir);
+    std::filesystem::create_directory(destDir);
+
+    writeConfig(jsonData);
+    sdbusplus::async::context ctx;
+
+    data_sync::Manager manager{ctx, std::move(extDataIface),
+                               ManagerTest::dataSyncCfgDir};
+
+    // Watch for dest path data change
+    data_sync::watch::inotify::DataWatcher dataWatcher(ctx, IN_NONBLOCK,
+                                                       IN_CREATE, destDir);
+    ctx.spawn(dataWatcher.onDataChange() |
+              sdbusplus::async::execution::then(
+                  [&destDir]([[maybe_unused]] const auto& dataOps) {
+        std::string destSubDir = destDir + "/Test/";
+        EXPECT_TRUE(std::filesystem::exists(destSubDir));
+    }));
+
+    // Write data after 1s so that the background sync events will be ready
+    // to catch.
+    ctx.spawn(sdbusplus::async::sleep_for(ctx, 1s) |
+              sdbusplus::async::execution::then([&ctx, &srcDir]() {
+        std::filesystem::create_directory(srcDir + "/Test");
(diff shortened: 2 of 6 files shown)


RameshIyyar:
add to approvelist

src/manager.cpp
```diff
@@ -151,7 +151,7 @@ sdbusplus::async::task<bool>
     Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
 {
     using namespace std::string_literals;
-    std::string syncCmd{"rsync --archive --compress"};
+    std::string syncCmd{"rsync --archive --compress --delete"};
```
aryakpadman: Regarding commit message :

>When syncing from BMC0 to BMC1, File b should be deleted from BMC1.

In example if DEST has file a and b , and SRC doesn't have both a and b, then both File a and b will get deleted , right ?

riya461: If the destination has multiple files that are not present in the source, while syncing the directories, each non-existing file in source will be deleted

test/full_sync_test.cpp
```diff
@@ -546,12 +546,14 @@ TEST_F(ManagerTest, FullSyncFailed)
...
-          {{"Path", ManagerTest::tmpDataSyncDataDir.string() + "/srcFile4"},
+          {{"Path",
+            ManagerTest::tmpDataSyncDataDir.string() + "/test/srcFile4"},
            {"DestinationPath",
-            ManagerTest::tmpDataSyncDataDir.string() + "/destFile4"},
+            ManagerTest::tmpDataSyncDataDir.string() + "/test/destFile4"},
            {"Description", "FullSync from Passive to Active bmc"},
            {"SyncDirection", "Passive2Active"},
-           {"SyncType", "Immediate"}}}}};
+           {"SyncType", "Periodic"},
```
RameshIyyar, line 555: why are we changing into different sync type now?

test/full_sync_test.cpp
```diff
@@ -582,8 +584,8 @@ TEST_F(ManagerTest, FullSyncFailed)
     ASSERT_EQ(ManagerTest::readData(srcFile3), data3);
 
     // Commented out the writing and verification for srcFile4 to simulate a
-    // failure scenario, where the source file at
-    // "/tmp/pdsDataDirXXXXXX/srcFile4" does not exist. This causes the rsync
+    // failure scenario, where the parent of source file srcFile4
```
RameshIyyar, line 587: >     // failure scenario, where the parent of source file srcFile4
>   // "/tmp/pdsDataDirXXXXXX/test/" does not exist. This causes the rsync

    // failure scenario, where the parent (/tmp/pdsDataDirXXXXXX/test) of source file "srcFile4"
    // does not exist. This causes the rsync

test/immediate_sync_test.cpp
```diff
@@ -177,6 +177,86 @@ TEST_F(ManagerTest, testDataDeleteInDir)
...
+    // Watch for dest path data change
+    data_sync::watch::inotify::DataWatcher dataWatcher(
+        ctx, IN_NONBLOCK, IN_DELETE_SELF, destPath);
+    ctx.spawn(dataWatcher.onDataChange() |
+              sdbusplus::async::execution::then(
+                  [&destPath]([[maybe_unused]] const auto& dataOps) {
+        EXPECT_FALSE(std::filesystem::exists(destPath));
+    }));
+
+    // Write data after 1s so that the background sync events will be ready
```
RameshIyyar, line 248: > Write data

Remove source file?

test/immediate_sync_test.cpp
```diff
@@ -177,6 +177,86 @@ TEST_F(ManagerTest, testDataDeleteInDir)
     ctx.run();
 }
```
RameshIyyar, line 179: It would be great if added a test case periodic sync.

src/data_watcher.cpp
```diff
@@ -240,14 +240,14 @@ std::optional<DataOperation>
...
-        _watchDescriptors.at(std::get<WD>(receivedEventInfo)) /
-        std::get<BaseName>(receivedEventInfo);
-
     // Process IN_CREATE only for DIR and skip for files as
     // all the file events are handled using IN_CLOSE_WRITE
     if ((std::get<EventMask>(receivedEventInfo) & IN_ISDIR) != 0)
     {
+        fs::path absCreatedPath =
+            _watchDescriptors.at(std::get<WD>(receivedEventInfo)) /
+            std::get<BaseName>(receivedEventInfo) / fs::path("");
```
RameshIyyar, line 249: Why do we need to create a empty path to append backslash?

cannot we simply append? 

For example,

```
fs::path absCreatedPath =
            (_watchDescriptors.at(std::get<WD>(receivedEventInfo)) /
            std::get<BaseName>(receivedEventInfo)).append(fs::path::preferred_separator);
```

test/full_sync_test.cpp
riya461: Here we are trying to failFullSync through an rsync failure. Previously, the case used was that the file was not present in the destination, but `--delete-missing-args` deals with it. So we are making it into the parent directory missing from destination case. There was a `writeData` that would trigger the `inotify`, but we have changed that to exclude this file now.

So any sync type would work. Can revert back to Immediate.

src/data_watcher.cpp
riya461: Should I add the case mention in `--delete` case, syncing two directories, handling creation, deletion and updation of sub-files/sub-directories?

test/full_sync_test.cpp
RameshIyyar: > So any sync type would work. Can revert back to Immediate.

Yes thats my point, we no need to change but, just want to check, is there any motive behind this?

src/data_watcher.cpp
RameshIyyar: Here, we are trying to add code to handle the IN_CREATE event for subdirectories. I noticed that there are test cases for other inotify events (IN_CLOSE_WRITE, IN_DELETE, IN_DELETE_SELF). So, similarly, should we add one for IN_CREATE?

Or are you saying that the existing test cases where a subdirectory is created while inotify is running? I believe all subdirectories are created before the watch starts (i,e before Manager instance created), meaning we are not simulating a scenario to catch the IN_CREATE event. Is that correct?

test/full_sync_test.cpp
riya461: Initially, `writeData` was initiated for all Immediate sync files so I changed, but now it is using  `writeData` only for files that are already present. It could be any type of sync now.

src/data_watcher.cpp
riya461: No, currently we do not have cases to handle IN_CREATE of subdirectory. 

So, the test case is to sync directories, and handling of creating a subdirectory right?

RameshIyyar: Yes

riya461: Have updated it in the [Config: Adding trailing / for directory path](https://github.com/ibm-openbmc/phosphor-data-sync/pull/20/commits/fdbff5cbd5a37b3d81aaa745e9c8369b6f18d151) commit

test/immediate_sync_test.cpp
riya461: Have updated it in the [rsync: Adding --delete option](https://github.com/ibm-openbmc/phosphor-data-sync/pull/20/commits/0d38159295b6769aa5b73218ce55134a1cb5af4a) commit

test/immediate_sync_test.cpp
```diff
@@ -346,3 +346,85 @@ TEST_F(ManagerTest, testDataChangeWhenSyncIsDisabled)
         << "The data should match with the data as the src was modified"
         << " and sync should take place at every modification.";
 }
+
+TEST_F(ManagerTest, testDataCreateInDir)
```
RameshIyyar, line 350: > testDataCreateInDir

Do you mean create a subdirectory? please correct it.

test/periodic_sync_test.cpp
```diff
@@ -352,3 +353,153 @@ TEST_F(ManagerTest, PeriodicDisablePropertyTest)
         << "The data should match with the data as 2.2s is passed"
         << " and sync should take place every 1s as per config.";
 }
+
+TEST_F(ManagerTest, PeriodicDataChangeInFile)
```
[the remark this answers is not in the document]
riya461: Will remove this test case, add one for `--delete-missing-args`

test/periodic_sync_test.cpp
```diff
@@ -352,3 +353,153 @@ TEST_F(ManagerTest, PeriodicDisablePropertyTest)
...
+    ManagerTest::writeData(destDirFile, destData);
+
+    ASSERT_EQ(ManagerTest::readData(srcDirFile), data);
+    ASSERT_EQ(ManagerTest::readData(destDirFile), destData);
+
+    data_sync::Manager manager{ctx, std::move(extDataIface),
+                               ManagerTest::dataSyncCfgDir};
+
+    // remove the file from srcDir
+    std::filesystem::remove(srcDirFile);
```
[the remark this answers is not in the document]
riya461: I will change the code to my understanding. 

I believe after waiting for fullsync, we should remove the file and then wait for deletion.

test/periodic_sync_test.cpp
```diff
@@ -1,3 +1,4 @@
+#include "data_watcher.hpp"
```
RameshIyyar, line 1: I think its not required

RameshIyyar (commented):
Just one minor comment

test/periodic_sync_test.cpp
riya461: Removed in latest push.

RameshIyyar (approved):

PR #22, opened 2025-03-24
Manager: Implement Sync Events Health property

harsh-agarwal1:
- This commit provides implementation of sync events health property.
- If sync is disabled, we mark sync events health as 'Paused'.
- When full sync completes successfully we mark sync events health as 'Ok'.
- If any of sync events fails, we mark sync events health as 'Critical'.
- Modified test cases.

Tested
- Verified on simics.

Change-Id: Ic99f090f1023ad49b9492410d64fcb19ebf75513

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -166,8 +166,11 @@ sdbusplus::async::task<bool>
     int result = std::system(syncCmd.c_str()); // NOLINT
     if (result != 0)
     {
-        // TODO:
-        // Retry and create error log and disable redundancy if retry is failed.
+        // TODOs:
+        // 1. Retry based on rsync error code
+        // 2. Create error log and Disable redundancy if retry fails
+        // 3. Perform a callout
+        setSyncEventsHealth(SyncEventsHealth::Critical);
         lg2::error("Error syncing: {PATH}", "PATH", dataSyncCfg._path);
 
         co_return false;
@@ -300,6 +303,7 @@ sdbusplus::async::task<void> Manager::startFullSync()
                             [](const auto& result) { return result; }))
     {
         _syncBMCDataIface.full_sync_status(FullSyncStatus::FullSyncCompleted);
+        setSyncEventsHealth(SyncEventsHealth::Ok);
         lg2::info("Full Sync completed successfully");
     }
     else
--- a/src/sync_bmc_data_ifaces.cpp
+++ b/src/sync_bmc_data_ifaces.cpp
@@ -62,6 +62,11 @@ bool SyncBMCDataIface::set_property([[maybe_unused]] disable_sync_t type,
     }
     disable_sync_ = disable;
     _manager.disableSyncPropChanged(disable);
+    if (sync_events_health_ != SyncEventsHealth::Critical)
+    {
+        _manager.setSyncEventsHealth(disable ? SyncEventsHealth::Paused
+                                             : SyncEventsHealth::Ok);
+    }
     return true;
 }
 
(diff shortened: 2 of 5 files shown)


src/manager.cpp
```diff
@@ -165,8 +165,11 @@ sdbusplus::async::task<bool>
     int result = std::system(syncCmd.c_str()); // NOLINT
     if (result != 0)
     {
-        // TODO:
-        // Retry and create error log and disable redundancy if retry is failed.
+        // TODOs:
+        // 1. Retry based on rsync error code
+        // 2. Create error log and Disable redundancy if retry fails
+        // 3. Perform a callout
+        _syncBMCDataIface.sync_events_health(SyncEventsHealth::Critical);
```
spinler, line 172: It's possible a sync could fail if someone reboots the passive BMC and a sync is attempted or in progress before we notice the reboot and disable sync.

When it comes back we'll just enable sync and do a full sync.  

I assume the sync health being in Critical until then is OK, right?

harsh-agarwal1: Yeah, as soon as passive BMC comes live again, we do a full sync, on successful completion of full sync, we will mark the sync health as 'Ok'.

spinler (approved):

src/sync_bmc_data_ifaces.cpp
```diff
@@ -54,6 +54,10 @@ sdbusplus::async::task<>
 bool SyncBMCDataIface::set_property([[maybe_unused]] disable_sync_t type,
                                     bool disable)
 {
+    if (sync_events_health_ == SyncEventsHealth::Ok && disable)
```
RameshIyyar, line 57: I think we could move line 67 to avoid checking every time the system or user tries to update the disable property.

Also, if the sync event health is Paused and sync is enabled, we should change it to Ok, right? From what I see, it currently only changes after a full sync, but we don’t need to wait for that here, do we? If the full sync fails, the health will be set to Critical anyway, right?

test/immediate_sync_test.cpp
```diff
@@ -148,6 +150,8 @@ TEST_F(ManagerTest, testDataChangeWhenSyncIsDisabled)
 
     std::string dataToWrite{"Data is modified"};
     std::string dataToStopEvent{"Close spawned inotify event."};
+    EXPECT_EQ(manager.getSyncEventsHealth(), SyncEventsHealth::Paused)
```
RameshIyyar, line 153: We should have test case to check Paused back to OK, right?

test/full_sync_test.cpp
```diff
@@ -127,6 +129,9 @@ TEST_F(ManagerTest, FullSyncA2PTest)
     data_sync::Manager manager{ctx, std::move(extDataIface),
                                ManagerTest::dataSyncCfgDir};
 
+    // Setting the SyncEventsHealth to Critical to test status change after full
+    // sync completes.
+    manager.setSyncEventsHealth(SyncEventsHealth::Critical);
```
RameshIyyar, line 134: I'm not sure why we need to use the existing test case (I am not fan of that approach but it your wish). Please check whether it violates the current test responsibilities. If it doesn't, we could document the specific cases covered by this test.

The same applies to the pause-to-OK case.

manishtwari: from a functional standpoint these changes appear acceptable for me. as Harsh set the property based on the sync status, and updating Sync health accordingly seems necessary once sync will complete. However, as Ramesh mentioned, we should document this since this test case also covers additional scenarios.

RameshIyyar (approved):

PR #23, opened 2025-04-11
DataSync: Added DBus Serialization Utilities

harsh-agarwal1:
- A JSON file was chosen so it's easy to see all the values at once and to capture them all in a dump or error log.
- Stores and Restores DBUS properties: 'FullSyncStatus', 'SyncEventsHealth' and 'DisableSync'
- Writes to the JSON whenever there's a change to above DBus properties.
- Design is compatible with introduction of new DBus properties.

Tested
- Added test cases.
- Verified on simics. Example JSON file generated:
```bash
$ cat /var/lib/phosphor-data-sync/persistence/dbus_props.json
{
    "Disable": false,
    "FullSyncStatus": 2,
    "SyncEventsHealth": 2
}
```

Change-Id: Id6e2acb1b9b692e5925b9fcab61dec9eef6bbbff

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -266,10 +266,51 @@ void Manager::disableSyncPropChanged(bool disableSync)
     }
 }
 
+void Manager::setFullSyncStatus(const FullSyncStatus& fullSyncStatus)
+{
+    if (_syncBMCDataIface.full_sync_status() == fullSyncStatus)
+    {
+        return;
+    }
+    _syncBMCDataIface.full_sync_status(fullSyncStatus);
+
+    try
+    {
+        data_sync::persist::update(data_sync::persist::key::fullSyncStatus,
+                                   fullSyncStatus);
+    }
+    catch (const std::exception& e)
+    {
+        lg2::error(
+            "Error writing fullSyncStatus property to JSON file: {ERROR}",
+            "ERROR", e);
+    }
+}
+
+void Manager::setSyncEventsHealth(const SyncEventsHealth& syncEventsHealth)
+{
+    if (_syncBMCDataIface.sync_events_health() == syncEventsHealth)
+    {
+        return;
+    }
+    _syncBMCDataIface.sync_events_health(syncEventsHealth);
+    try
+    {
+        data_sync::persist::update(data_sync::persist::key::syncEventsHealth,
+                                   syncEventsHealth);
+    }
+    catch (const std::exception& e)
+    {
+        lg2::error(
+            "Error writing syncEventsHealth property to JSON file: {ERROR}",
+            "ERROR", e);
+    }
+}
+
 // NOLINTNEXTLINE
 sdbusplus::async::task<void> Manager::startFullSync()
 {
-    _syncBMCDataIface.full_sync_status(FullSyncStatus::FullSyncInProgress);
+    setFullSyncStatus(FullSyncStatus::FullSyncInProgress);
     lg2::info("Full Sync started");
 
     auto fullSyncStartTime = std::chrono::steady_clock::now();
@@ -308,7 +349,7 @@ sdbusplus::async::task<void> Manager::startFullSync()
     if (std::ranges::all_of(syncResults,
                             [](const auto& result) { return result; }))
     {
-        _syncBMCDataIface.full_sync_status(FullSyncStatus::FullSyncCompleted);
+        setFullSyncStatus(FullSyncStatus::FullSyncCompleted);
         setSyncEventsHealth(SyncEventsHealth::Ok);
         lg2::info("Full Sync completed successfully");
     }
@@ -317,11 +358,11 @@ sdbusplus::async::task<void> Manager::startFullSync()
         // Forcefully marking full sync as successful, even if data syncing
         // fails.
         // TODO: Revert this workaround once the proper logic is implemented
-        _syncBMCDataIface.full_sync_status(FullSyncStatus::FullSyncCompleted);
+        setFullSyncStatus(FullSyncStatus::FullSyncCompleted);
         setSyncEventsHealth(SyncEventsHealth::Ok);
         lg2::info("Full Sync passed temporarily despite sync failures");
 
-        // _syncBMCDataIface.full_sync_status(FullSyncStatus::FullSyncFailed);
+        // setFullSyncStatus(FullSyncStatus::FullSyncFailed);
         // lg2::info("Full Sync failed");
     }
 
--- a/src/manager.hpp
+++ b/src/manager.hpp
@@ -4,6 +4,7 @@
 
 #include "data_sync_config.hpp"
 #include "external_data_ifaces.hpp"
+#include "persistent.hpp"
 #include "sync_bmc_data_ifaces.hpp"
 
 #include <filesystem>
@@ -90,6 +91,13 @@ class Manager
     }
 
     /**
+     * @brief Helper API sets the full sync Dbus status-property.
+     *
+     * @param[in] fullSyncStatus - The Full Sync Status property being set.
+     */
+    void setFullSyncStatus(const FullSyncStatus& fullSyncStatus);
+
+    /**
      * @brief Helper API to start events when Disable sync property is changed.
      *        - If the Disable sync property is set to true, it stops all sync
      *          events. Otherwise, it starts all sync events.
@@ -124,10 +132,7 @@ class Manager
      * @param[in] syncEventsHealth - The sync events health property value being
      * set.
      */
-    void setSyncEventsHealth(const SyncEventsHealth& syncEventsHealth)
-    {
-        _syncBMCDataIface.sync_events_health(syncEventsHealth);
-    }
+    void setSyncEventsHealth(const SyncEventsHealth& syncEventsHealth);
 
   private:
     /**
--- a/src/persistent.hpp
+++ b/src/persistent.hpp
@@ -1,0 +1,106 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <nlohmann/json.hpp>
+
+#include <filesystem>
+#include <optional>
+
+namespace data_sync::persist
+{
+
+extern std::filesystem::path DBusPropDataFile;
+
+namespace key
+{
+constexpr auto disable = "Disable";
+constexpr auto fullSyncStatus = "FullSyncStatus";
+constexpr auto syncEventsHealth = "SyncEventsHealth";
+} // namespace key
+
+namespace util
+{
+
+/**
+ * @brief Helper function to update a JSON file
+ *
+ * @param[in] json - The JSON to update
+ * @param[in] path - The path to the file
+ */
+void writeFile(const nlohmann::json& json, const std::filesystem::path& path);
+
+} // namespace util
+
+/**
+ * @brief Function to read a JSON file
+ *
+ * @param[in] path - The path to the file
+ *
+ * @return optional<json> - The JSON data, or std::nullopt if it
+ *                          didn't exist or was corrupt.
+ */
+std::optional<nlohmann::json> readFile(const std::filesystem::path& path);
+
+/**
+ * @brief Updates "name": <value>  JSON to the file specified
+ *
+ * @tparam - The data type
+ * @param[in] name - The key to save the value under
+ * @param[in] value - The value to save
+ */
+template <typename T>
+void update(std::string_view name, const T& value,
+            const std::filesystem::path& path = DBusPropDataFile)
+{
+    auto json = readFile(path).value_or(nlohmann::json::object());
+    if constexpr (std::is_enum_v<T>)
+    {
+        json[name] = std::to_underlying(value);
+    }
+    else
+    {
+        json[name] = value;
+    }
+    util::writeFile(json, path);
+}
+
+/**
+ * @brief Reads the value of the key specified in the file specified
+ *        Specifically, for unit testing purposes.
+ *
+ * @tparam T - The data type
+ * @param[in] name - The key the value is saved under
+ * @param[in] path - The path to the file
+ *
+ * @return optional<T> - The value, or std::nullopt if the file or
+ *                       key isn't present.
+ */
+template <typename T>
+std::optional<T> read(std::string_view name,
+                      const std::filesystem::path& path = DBusPropDataFile)
+{
+    auto json = readFile(path);
+    if (!json)
+    {
+        return std::nullopt;
+    }
+
+    auto it = json->find(name);
+    if (it != json->end())
+    {
+        if constexpr (std::is_enum_v<T>)
+        {
+            auto value = it->get<std::underlying_type_t<T>>();
+            return static_cast<T>(value);
+        }
+        else
+        {
+            return it->get<T>();
+        }
+    }
+
+    return std::nullopt;
+}
+
+} // namespace data_sync::persist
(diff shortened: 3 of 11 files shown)


src/persistent_data.cpp
```diff
@@ -0,0 +1,50 @@
+#include "persistent_data.hpp"
```
RameshIyyar, line 1: Licence is missing "// SPDX-License-Identifier: Apache-2.0" - The same applies to all newly added files.

src/persistent_data.hpp
```diff
@@ -0,0 +1,104 @@
+#pragma once
+
+#include <nlohmann/json.hpp>
+
+#include <filesystem>
+#include <optional>
+
+namespace data
```
RameshIyyar, line 8: Just thoughts, cannot we make it as class?

src/sync_bmc_data_ifaces.cpp
```diff
@@ -18,9 +18,41 @@ SyncBMCDataIface::SyncBMCDataIface(sdbusplus::async::context& ctx,
         SyncBMCDataIface>(ctx, SyncBMCData::instance_path),
     _manager(manager), _ctx(ctx)
 {
+    loadPropertiesFromFile();
```
RameshIyyar, line 21: is this correct?

I think we will emit the signal while restoring from the persistent file, right?

src/sync_bmc_data_ifaces.cpp
```diff
@@ -67,6 +99,15 @@ bool SyncBMCDataIface::set_property([[maybe_unused]] disable_sync_t type,
         _manager.setSyncEventsHealth(disable ? SyncEventsHealth::Paused
                                              : SyncEventsHealth::Ok);
     }
+    try
+    {
+        data::write(data::key::disable, disable);
+    }
+    catch (const std::exception& e)
+    {
+        lg2::info("Could not serialize Disable value of {DISABLE}: {ERROR}",
```
RameshIyyar, line 108: > Disable

SyncEventsHealth?

test/persistent_data_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+    // should load the persistent data as defined.
+    data::write(data::key::fullSyncStatus, FullSyncStatus::FullSyncInProgress);
+    data::write(data::key::syncEventsHealth, SyncEventsHealth::Critical);
+    data_sync::Manager manager{ctx, std::move(extDataIface),
+                               ManagerTest::dataSyncCfgDir};
+
+    EXPECT_EQ(data::read<FullSyncStatus>(data::key::fullSyncStatus),
+              FullSyncStatus::FullSyncInProgress);
+    EXPECT_EQ(manager.getFullSyncStatus(), FullSyncStatus::FullSyncInProgress)
+        << "FullSyncInProgress was set to InProgress in persistent file.";
```
RameshIyyar, line 84: > was set

is not set?

test/persistent_data_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    EXPECT_EQ(data::read<SyncEventsHealth>(data::key::syncEventsHealth),
+              SyncEventsHealth::Critical);
+    EXPECT_EQ(manager.getSyncEventsHealth(), SyncEventsHealth::Critical)
+        << "SyncEventsHealth was set to Critical in persistent file.";
```
RameshIyyar, line 89: > was set

is not set?

src/sync_bmc_data_ifaces.cpp
RameshIyyar: I think it will work, since we claim name after this. please check it.

src/persistent_data.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * @brief Writes "name": <value>  JSON to the file specified
+ *
+ * @tparam - The data type
+ * @param[in] name - The key to save the value under
+ * @param[in] value - The value to save
+ */
+template <typename T>
+void write(std::string_view name, const T& value,
```
aryakpadman, line 51: It looks like this API isn't directly writing data to the JSON file. Instead, it's constructing the updated JSON and delegating the actual write operation to util::writeFile. 

Given that, would naming it `updatePersistentFile` be more appropriate?

src/sync_bmc_data_ifaces.hpp
```diff
@@ -43,6 +43,12 @@ class SyncBMCDataIface :
     SyncBMCDataIface(sdbusplus::async::context& ctx,
                      data_sync::Manager& manager);
 
+    /**
+     * @brief Reads from the persistence file and sets the property values if
+     * available.
+     */
+    void loadPropertiesFromFile();
```
aryakpadman, line 50: can we rename as `restoreDBusProperties` ?

src/persistent_data.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include "phosphor-logging/lg2.hpp"
+
+#include <format>
+#include <fstream>
+
+namespace data
```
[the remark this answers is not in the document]
harsh-agarwal1: Made changes according to Ramesh comments, now it looks like:` persist::read()`

src/persistent_data.cpp
```diff
@@ -0,0 +1,50 @@
...
+{
+
+std::optional<nlohmann::json> readFile(const std::filesystem::path& path)
+{
+    if (std::filesystem::exists(path))
+    {
+        std::ifstream stream{path};
+        try
+        {
+            return nlohmann::json::parse(stream, nullptr, true);
```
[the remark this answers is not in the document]
harsh-agarwal1: we enabled the strict mode with the 3rd argument.

src/sync_bmc_data_ifaces.cpp
harsh-agarwal1: We won't be emitting signal while restoring from persistent file. Tested this on simics. 
Used command: `busctl monitor --match 'sender=xyz.openbmc_project.Control.SyncBMCData,path=/xyz/openbmc_project/control/sync_bmc_data,interface=org.freedesktop.DBus.Properties,type=signal'`

src/sync_bmc_data_ifaces.cpp
harsh-agarwal1: We are changing disable property. Checking `SyncEventsHealth` property for our logic. Hence Disable in the log message.

src/persistent_data.hpp
harsh-agarwal1: Inspired from state-manager code, felt it would keep code simple and would help to upstream ([commit link](https://github.com/ibm-openbmc/phosphor-state-manager/commit/49870770764c7282c24e487ed1974e65b112f5d8)).

spinler: I was thinking of changing mine to a class since it isn't testable in the real code as-is since all the functions default to a BMC only path.  I see you got around that by changing the global path.  Not implying you need to change anything, it's up to you.

src/sync_bmc_data_ifaces.cpp
```diff
@@ -18,9 +16,42 @@ SyncBMCDataIface::SyncBMCDataIface(sdbusplus::async::context& ctx,
...
+        if (auto it = json->find(persist::key::syncEventsHealth);
+            it != json->end())
+        {
+            sync_events_health_ = static_cast<SyncEventsHealth>(
+                it->get<std::underlying_type_t<SyncEventsHealth>>());
+        }
+    }
+    catch (const std::exception& e)
+    {
+        lg2::error("Failed trying to obtain previous value of DBus properties");
```
spinler, line 51: I know this is unlikely to fail, but you may want to print the e.what() here in case it ever does.

test/persistent_data_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+std::filesystem::path ManagerTest::dataSyncCfgDir;
+std::filesystem::path ManagerTest::tmpDataSyncDataDir;
+
+using FullSyncStatus = sdbusplus::common::xyz::openbmc_project::control::
+    SyncBMCData::FullSyncStatus;
+using SyncEventsHealth = sdbusplus::common::xyz::openbmc_project::control::
+    SyncBMCData::SyncEventsHealth;
+
+TEST_F(ManagerTest, testDBusDataPersistency)
```
spinler, line 12: Wouldn't this belong in manager_test.cpp?

src/sync_bmc_data_ifaces.cpp
```diff
@@ -18,9 +16,42 @@ SyncBMCDataIface::SyncBMCDataIface(sdbusplus::async::context& ctx,
...
 {
+    restoreDBusProperties();
     emit_added();
 }
 
+void SyncBMCDataIface::restoreDBusProperties()
+{
+    try
+    {
+        auto json = persist::util::readFile(persist::DBusPropDataFile);
```
spinler, line 27: Since you're using readFile outside of persist.hpp, you could probably remove it from the util namespace to indicate it can be used by anyone.

test/persistent_data_test.cpp
harsh-agarwal1: `manager_test.cpp` have test cases which requires testing of common config JSON, that would interfere with our use case, hence wrote tests in a new file.

src/sync_bmc_data_ifaces.cpp
harsh-agarwal1: Done

src/sync_bmc_data_ifaces.cpp
harsh-agarwal1: Done

src/persistent_data.hpp
harsh-agarwal1: Changed to update, now calls looks like: `persist::update()`

src/sync_bmc_data_ifaces.hpp
harsh-agarwal1: Done

src/persistent_data.hpp
harsh-agarwal1: Sure!

src/persistent.hpp
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @param[in] json - The JSON to update
+ * @param[in] path - The path to the file
+ */
+void writeFile(const nlohmann::json& json, const std::filesystem::path& path);
+
+} // namespace util
+
+/**
+ * @brief Helper function to read a JSON file
```
spinler, line 36: nit: Not really a helper function anymore.

test/persistent_data_test.cpp
spinler: I think as a first testcase it would be good to have one that purely exercises the new persist interfaces by themselves.

src/persistent.hpp
harsh-agarwal1: Done!

test/persistent_data_test.cpp
harsh-agarwal1: Created new test case `testReadWritePersistencyFile` for same.

aryakpadman (approved):

manishtwari (approved):

src/persistent_data.cpp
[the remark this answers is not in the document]
RameshIyyar: I agree with @aryakpadman...All namespace should be under `data_sync` which is meant for the data sync repo so it should be like `data_sync::persist`

src/persistent_data.hpp
RameshIyyar: I would prefer class but its up to you.

src/sync_bmc_data_ifaces.cpp
RameshIyyar: Capture DBus property name in the trace.

test/persistent_data_test.cpp
RameshIyyar: I agree with @spinler. This test case functions more like a flow test—it involves synchronizing data and setting properties that need to persist. As such, it would be more appropriate to include it in the manager test suite.

> requires testing of common config JSON

I could not understand, I remember we removed the common JSON config dependency, every test case has own config JSON. could you help me to understand?

src/manager.hpp
```diff
@@ -89,6 +92,33 @@ class Manager
...
     }
 
+    /**
+     * @brief Helper API sets the full sync Dbus status-property.
+     *
+     * @param[in] fullSyncStatus - The Full Sync Status property being set.
+     */
+    void setFullSyncStatus(const FullSyncStatus& fullSyncStatus)
+    {
+        if (getFullSyncStatus() == fullSyncStatus)
```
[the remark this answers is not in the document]
RameshIyyar: I think I suggested to use `_syncBMCDataIface.full_sync_status` return value a while ago. we dont need one more function for this.

test/manager_test.cpp
```diff
@@ -77,4 +82,192 @@ TEST_F(ManagerTest, ParseDataSyncCfg)
 
     EXPECT_TRUE(manager.containsDataSyncCfg(data_sync::config::DataSyncConfig(
         ManagerTest::commonJsonData["Files"][0], false)));
+    std::filesystem::remove(dataSyncCommonCfgFile);
```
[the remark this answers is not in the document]
harsh-agarwal1: Used `writeConfig()` to use the helper APIs.

src/persistent_data.cpp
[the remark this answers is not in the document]
harsh-agarwal1: Done

test/manager_test.cpp
```diff
@@ -78,3 +77,197 @@ TEST_F(ManagerTest, ParseDataSyncCfg)
     EXPECT_TRUE(manager.containsDataSyncCfg(data_sync::config::DataSyncConfig(
         ManagerTest::commonJsonData["Files"][0], false)));
 }
+
+TEST_F(ManagerTest, testReadWritePersistencyFile)
```
spinler, line 81: This test would be best in persist_test.cpp since it just tests that code.

harsh-agarwal1: Done

test/persistent_data_test.cpp
harsh-agarwal1: Used `writeConfig()` to write the config file, eliminating the use of creation and deletion of new config file.

src/manager.cpp
```diff
@@ -266,10 +266,55 @@ void Manager::disableSyncPropChanged(bool disableSync)
     }
 }
 
+void Manager::setFullSyncStatus(const FullSyncStatus& fullSyncStatus)
+{
+    if (_syncBMCDataIface.full_sync_status() == fullSyncStatus)
```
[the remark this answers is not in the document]
harsh-agarwal1: `full_sync_status()` returns a void, the build code has function definition as: `void full_sync_status(Arg&& new_value)`.

src/manager.cpp
```diff
@@ -266,10 +266,55 @@ void Manager::disableSyncPropChanged(bool disableSync)
...
+    {
+        lg2::error(
+            "Error writing fullSyncStatus property to JSON file: {ERROR}",
+            "ERROR", e);
+    }
+}
+
+void Manager::setSyncEventsHealth(const SyncEventsHealth& syncEventsHealth)
+{
+    if (_syncBMCDataIface.sync_events_health() == syncEventsHealth)
```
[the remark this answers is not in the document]
harsh-agarwal1: `sync_events_health()` returns void as per generated code `void sync_events_health(Arg&& new_value)`.

src/manager.cpp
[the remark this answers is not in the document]
RameshIyyar: ah ok...Please remove trace.

PR #21, opened 2025-03-17
DataSync: Enable Sibling BMC Synchronization with Stunnel and Rsyncd

RameshIyyar:

PR #26, opened 2025-06-04
Enable the necessary rsync CLI options

aryakpadman:
This PR is dependent on the changes from [PR21](https://github.com/ibm-openbmc/phosphor-data-sync/pull/21) and is not merged yet.

Hence please review the below mentioned 2 commits as part of this PR :

1. [rsync-cli : Add --relative flag](https://github.com/ibm-openbmc/phosphor-data-sync/commit/59d18543ffc1d2d2432fa85e4f2f08be1775cb64)
2. [rsync-cli : Adding the necessary rsync flags](https://github.com/ibm-openbmc/phosphor-data-sync/commit/3299835bad5ec8e4a2aa7c39ad1c1303c727d728)

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -152,7 +152,8 @@ sdbusplus::async::task<bool>
 {
     using namespace std::string_literals;
     std::string syncCmd{
-        "rsync --archive --compress --delete --delete-missing-args"};
+        "rsync --archive --compress --relative --delete --delete-missing-args "};
+
     syncCmd.append(" "s + dataSyncCfg._path);
 
 #ifdef UNIT_TEST
@@ -161,8 +162,9 @@ sdbusplus::async::task<bool>
     // TODO Support for remote (i,e sibling BMC) copying needs to be added.
 #endif
 
-    // Add destination data path
-    syncCmd.append(dataSyncCfg._destPath.value_or(dataSyncCfg._path));
+    // Add destination data path if configured
+    syncCmd.append(dataSyncCfg._destPath.value_or(fs::path("")));
+
     int result = std::system(syncCmd.c_str()); // NOLINT
     if (result != 0)
     {
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -151,8 +151,12 @@ sdbusplus::async::task<bool>
     Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
 {
     using namespace std::string_literals;
+
+    // For more details about CLI options, refer rsync man page.
+    // https://download.samba.org/pub/rsync/rsync.1#OPTION_SUMMARY
     std::string syncCmd{
-        "rsync --archive --compress --relative --delete --delete-missing-args "};
+        "rsync --compress --recursive --perms --group --owner --times --atimes"
+        " --update --relative --delete --delete-missing-args "};
 
     syncCmd.append(" "s + dataSyncCfg._path);
 
--- a/test/full_sync_test.cpp
+++ b/test/full_sync_test.cpp
@@ -2,6 +2,8 @@
 
 #include "manager_test.hpp"
 
+namespace fs = std::filesystem;
+
 std::filesystem::path ManagerTest::dataSyncCfgDir;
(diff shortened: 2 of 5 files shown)


src/manager.cpp
```diff
@@ -154,7 +154,8 @@ sdbusplus::async::task<bool>
 {
     using namespace std::string_literals;
     std::string syncCmd{
-        "rsync --archive --compress --relative --delete --delete-missing-args "};
+        "rsync --compress --recursive --perms --group --owner --times --atimes"
+        " --update --relative --delete --delete-missing-args "};
```
RameshIyyar, line 158: Commit message:

>  - Adding `--update` flag which forces rsync to skip any files which
   exist on the destination and have a modified time that is newer
   than the source file.

This is a key point—we need to clearly mention the reason for adding this flag. The remaining flags are self-explanatory. We specifically included the --update flag to support bidirectional sync, as there's a possibility that the destination might have newer data. Without this flag, rsync could overwrite the newer files with older versions from the source. To avoid that, we're using --update.

spinler:
In the commit message for "rsync-cli : Add --relative flag", can you explain why it's necessary?

aryakpadman:
> > This PR is dependent on the changes from #21 and is not merged yet.
> 
> Is that because of a merge conflict, or just functionally rsync doesn't work without that?



> > This PR is dependent on the changes from #21 and is not merged yet.
> 
> Is that because of a merge conflict, or just functionally rsync doesn't work without that?

Yes the PR21 has changes which enables mTLS via stunnel and rsyncd module name change etc which are required for the functionality to work.

The commit [rsync-cli : Adding the necessary rsync flags](https://github.com/ibm-openbmc/phosphor-data-sync/commit/3299835bad5ec8e4a2aa7c39ad1c1303c727d728) will not break if PR21 is not there, but the changes are required for the commit [rsync-cli : Add --relative flag](https://github.com/ibm-openbmc/phosphor-data-sync/commit/59d18543ffc1d2d2432fa85e4f2f08be1775cb64)

src/manager.cpp
```diff
@@ -169,9 +169,8 @@ sdbusplus::async::task<bool>
     syncCmd.append(rsyncdURL);
 #endif
 
-    // Add destination data path
-    syncCmd.append(dataSyncCfg._destPath.value_or(dataSyncCfg._path).string());
-    // syncCmd.append(dataSyncCfg._destPath.value_or(fs::path("")));
+    // Add destination data path if configured
+    syncCmd.append(dataSyncCfg._destPath.value_or(fs::path("")));
```
manishtwari, line 174: with earlier change it ensured rsync always had a valid destination even if DestinationPath was missing, with this change `dataSyncCfg._destPath` is empty if it is not configured. 
Just confirming: do we guarantee DestinationPath is always set?

src/manager.cpp
harsh-agarwal1, line 158: Which option do --archive inheritably have which we are omitting?

harsh-agarwal1 (approved):

src/manager.cpp
aryakpadman: We are removing the --links option , which was part of --archive. 

The option is to add symlinks to the transferred files instead of noisily ignoring them with a "non-regular file" warning for each symlink encountered, which we're not planning to support.

aryakpadman: Updated the commit message also.

riya461 (approved):

src/manager.cpp
```diff
@@ -161,8 +162,9 @@ sdbusplus::async::task<bool>
     // TODO Support for remote (i,e sibling BMC) copying needs to be added.
 #endif
 
-    // Add destination data path
-    syncCmd.append(dataSyncCfg._destPath.value_or(dataSyncCfg._path));
+    // Add destination data path if configured
+    syncCmd.append(dataSyncCfg._destPath.value_or(fs::path("")));
```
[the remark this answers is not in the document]
aryakpadman: Done.

harsh-agarwal1 (approved):

manishtwari (approved):

spinler (approved):

PR #27, opened 2025-06-05
Add the support to handle IN_MOVED_TO and IN_MOVED_FROM inotify events

aryakpadman:
This PR commits is dependent on PR21 and PR26 , which are yet to merge and contains the commits from PR21 and PR26 also. Hence review the below mentioned commits.

1. [inotify : process the IN_MOVED_TO signal](https://github.com/ibm-openbmc/phosphor-data-sync/commit/3adad6cce423c2ec2dd3696305897aa12c1add9d)
2. [inotify : Process IN_MOVED_FROM signal](https://github.com/ibm-openbmc/phosphor-data-sync/commit/feec13e889cd496a1d1c28f4369b452e89fa50eb)
3. [inotify : Watch for IN_MOVED_FROM and IN_MOVED_TO signals](https://github.com/ibm-openbmc/phosphor-data-sync/commit/8bf67403377ec8c3a79969b3f2e201de0b3f0017)

### DIFF
--- a/src/data_watcher.cpp
+++ b/src/data_watcher.cpp
@@ -167,7 +167,7 @@ std::optional<std::vector<EventInfo>> DataWatcher::readEvents()
         auto* receivedEvent = reinterpret_cast<inotify_event*>(&buffer[offset]);
 
         receivedEvents.emplace_back(receivedEvent->wd, receivedEvent->name,
-                                    receivedEvent->mask);
+                                    receivedEvent->mask, receivedEvent->cookie);
 
         offset += offsetof(inotify_event, name) + receivedEvent->len;
     }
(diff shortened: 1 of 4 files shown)


src/data_watcher.cpp
```diff
@@ -323,6 +327,24 @@ std::optional<DataOperation>
...
+    // Case 1 : If a file inside a configured and watching directory is renamed.
+    // Case 2 : A file is moved to a configured and watching directory.
+    // Case 3 : If a configured and watching file itself is renamed.
+    fs::path absCopiedPath =
+        _watchDescriptors.at(std::get<WD>(receivedEventInfo)) /
+        std::get<BaseName>(receivedEventInfo);
+
+    if (absCopiedPath.string().starts_with(_dataPathToWatch.string()))
+    {
+        lg2::debug("Processing IN_MOVED_TO for {PATH}", "PATH", absCopiedPath);
```
RameshIyyar, line 364: Please include cookie so that we could associated IN_MOVED_TO/FROM event while debugging.

We will see a bunch of debug traces during PEL archive.

src/data_watcher.cpp
```diff
@@ -191,6 +191,10 @@ std::optional<DataOperation>
          */
         return processCreate(receivedEventInfo);
     }
+    else if ((std::get<EventMask>(receivedEventInfo) & IN_MOVED_TO) != 0)
```
RameshIyyar, line 198: Commit Message:

Please include why this event is required to process in the commit message.

RameshIyyar, line 198: Commit Message:

Please include why this event is required to process in the commit message.

src/data_watcher.cpp
RameshIyyar, line 364: Please include cookie so that we could associated IN_MOVED_TO/FROM event while debugging.

We will see a bunch of debug traces during PEL archive.

src/data_watcher.cpp
```diff
@@ -323,6 +327,24 @@ std::optional<DataOperation>
...
+    // Case 2 : A file is moved to a configured and watching directory.
+    // Case 3 : If a configured and watching file itself is renamed.
+    fs::path absCopiedPath =
+        _watchDescriptors.at(std::get<WD>(receivedEventInfo)) /
+        std::get<BaseName>(receivedEventInfo);
+
+    if (absCopiedPath.string().starts_with(_dataPathToWatch.string()))
+    {
+        lg2::debug("Processing IN_MOVED_TO for {PATH}", "PATH", absCopiedPath);
+        return std::make_pair(absCopiedPath, DataOps::COPY);
```
RameshIyyar, line 365: I feel, COPY is not making sense here but, Do we need to maintain the DataOperation enumeration, even for future use? Since we're now directly adding all rsync options, I'm wondering if it's worth continuing to maintain DataOperation in the code.

We could remove later too if its unused.

test/immediate_sync_test.cpp
```diff
@@ -413,3 +413,108 @@ TEST_F(ManagerTest, testDataCreateInSubDir)
...
+    EXPECT_CALL(*mockExtDataIfaces, fetchSiblingBmcPos())
+        // NOLINTNEXTLINE
+        .WillRepeatedly([]() -> sdbusplus::async::task<> { co_return; });
+
+    nlohmann::json jsonData = {
+        {"Directories",
+         {{{"Path", ManagerTest::tmpDataSyncDataDir.string() + "/srcDir1/"},
+           {"DestinationPath",
+            ManagerTest::tmpDataSyncDataDir.string() + "/destDir1/"},
+           {"Description", "Directory to test immediate sync on file deletion"},
```
RameshIyyar, line 444: > file deletion

file move?

the same comment applies to other places.

src/data_watcher.cpp
aryakpadman: Yes, agreed. Initially, the enum was introduced in the immediate sync logic to help identify deleted files. However, upon deeper investigation, it is figured out that rsync already provides a mechanism to detect deleted files using the `--stat` option. This approach is more robust as it can gather the info during full and periodic sync too.

Is it fine If I treat this update as a separate cleanup commit?

src/data_watcher.cpp
aryakpadman: Addressed.

harsh-agarwal1 (approved):

manishtwari (approved):

aryakpadman:
> @aryakpadman Please rebase on top of latest main instead of pushing on top of mTLS PR...it takes sometime...

Rebased on top of latest PR.

RameshIyyar (approved):

spinler (approved):

PR #28, opened 2025-06-09
Dev/retry

manishtwari:


src/manager.hpp
```diff
@@ -178,14 +178,16 @@ class Manager
...
      *
      * @return Returns true if sync succeeds; otherwise, returns false
      *
      */
     sdbusplus::async::task<bool>
         syncData(const config::DataSyncConfig& dataSyncCfg,
                  const std::string srcPath = "",
-                 const std::string destPath = "");
+                 const std::string destPath = "",
+                 int retryCount = 0);
```
spinler, line 190: nit: usually size_t is used for data for counts.

src/manager.cpp
```diff
@@ -688,21 +700,56 @@ sdbusplus::async::task<bool>
     lg2::debug("Elapsed time for sync: [{DURATION_SECONDS}] seconds for {CMD}",
               "DURATION_SECONDS", syncElapsedTime.count(), "CMD", syncCmd);
 
-    if (ret.first != 0)
+    if (ret.first == 0)
+    {
+        co_return true;
+    }
+
+    const std::string srcLog = srcPath.empty() ? dataSyncCfg._path.string()
```
spinler, line 708: nit: this function is already using auto, e.g. lines 697,698, so probably appropriate here as well.

src/manager.cpp
```diff
@@ -543,8 +543,20 @@ sdbusplus::async::task<bool>
     // NOLINTNEXTLINE
     Manager::syncData(const config::DataSyncConfig& dataSyncCfg,
                       const std::string srcPath,
-                      const std::string destPath)
+                      const std::string destPath,
+                      size_t retryCount)
 {
+    constexpr size_t defaultMaxAttempts = 1;
```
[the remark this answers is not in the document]
manishtwari: addressed

test/data_sync_config_test.cpp
```diff
@@ -278,8 +278,8 @@ TEST(DataSyncConfigParserTest, TestFileSyncWithInvalidSyncType)
     EXPECT_EQ(dataSyncConfig._syncType, data_sync::config::SyncType::Immediate);
     EXPECT_EQ(dataSyncConfig._periodicityInSec, std::nullopt);
     EXPECT_EQ(dataSyncConfig._retry, std::nullopt);
```
riya461, line 280: Here we are checking the equality of retry value as `nullopt`, but according to the code change in [Line:76](https://github.com/ibm-openbmc/phosphor-data-sync/pull/28/files#diff-81eda4aee1c413401d50cd5b542476127f7dba8b32b673b23203f6ac722781adR76) - data_sync_config.cpp, a default value is being assigned, which will fail the UT.

RameshIyyar:
add to approvelist

PR #30, opened 2025-06-16
DataSync: Added signal-triggered watched paths logging utility

harsh-agarwal1:
- Introduced a mechanism to handle SIGTRAP signal for debugging purposes.
- Added support to read a list of paths from a predefined file and log, via journalctl, whether each path is currently being watched or not.
- On receiving the SIGTRAP signal, the system logs all currently watched paths to a file for inspection.
- Helps in live debugging and verification of inotify watch status without restarting services.

Tested
- Added test cases.
- Verified on simics. Example JSON file generated:
```bash
$ cat /tmp/current_watched_paths.json
[
    "/var/lib",
    "/var/lib/phosphor-state-manager"
]
```
Change-Id: I251fcf15c5307a2ceca5e464d0bd2ba1b98f7974

### DIFF
--- a/src/data_watcher.cpp
+++ b/src/data_watcher.cpp
@@ -20,6 +20,8 @@ DataWatcher::DataWatcher(sdbusplus::async::context& ctx, const int inotifyFlags,
         std::make_unique<sdbusplus::async::fdio>(ctx, _inotifyFileDescriptor()))
 {
     createWatchers(_dataPathToWatch);
+    std::lock_guard<std::mutex> lock(getMutex());
+    getAllWatchers().push_back(this);
 }
 
 DataWatcher::~DataWatcher()
@@ -33,6 +35,9 @@ DataWatcher::~DataWatcher()
             }
         });
     }
+    auto& instances = getAllWatchers();
+    instances.erase(std::remove(instances.begin(), instances.end(), this),
+                    instances.end());
 }
 
 int DataWatcher::inotifyInit() const
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -7,6 +7,7 @@
 #include <nlohmann/json.hpp>
 #include <phosphor-logging/lg2.hpp>
 
+#include <csignal>
 #include <cstdlib>
 #include <exception>
 #include <fstream>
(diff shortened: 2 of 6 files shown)


src/manager.cpp
```diff
@@ -46,9 +47,74 @@ sdbusplus::async::task<> Manager::init()
...
     co_return co_await startSyncEvents();
 }
 
+// NOLINTNEXTLINE
+sdbusplus::async::task<> Manager::monitorSignalForDebug()
+{
+    while (!_ctx.stop_requested())
+    {
+        co_await sdbusplus::async::sleep_for(_ctx, std::chrono::seconds(1));
+        std::signal(SIGTRAP, logSignalHandler);
```
spinler, line 60: Other services use SIGUSR1 to dump data.

Also, why do you need to install a signal handler more than once?

test/manager_test.cpp
```diff
@@ -204,3 +206,117 @@ TEST_F(ManagerTest, testDBusDataPersistency)
...
+    EXPECT_CALL(*mockExtDataIfaces, fetchRbmcCredentials())
+        // NOLINTNEXTLINE
+        .WillRepeatedly([]() -> sdbusplus::async::task<> { co_return; });
+
+    nlohmann::json jsonData = {
+        {"Files",
+         {
+             {{"Path", ManagerTest::tmpDataSyncDataDir.string() + "/srcFile1"},
+              {"DestinationPath",
+               ManagerTest::tmpDataSyncDataDir.string() + "/destFile1"},
```
aryakpadman, line 242: I am enabling `--relative` flag in RSYNC CLI as part of one of PR for rsync CLI options. This will break the test cases in this pattern if the PR-26 merged before this one.

In that case , you can refer the PR where I fixed the existing test cases : https://github.com/ibm-openbmc/phosphor-data-sync/pull/26

src/data_watcher.cpp
```diff
@@ -18,6 +18,7 @@ DataWatcher::DataWatcher(sdbusplus::async::context& ctx, const int inotifyFlags,
     _fdioInstance(
         std::make_unique<sdbusplus::async::fdio>(ctx, _inotifyFileDescriptor()))
 {
+    getAllWatchers().push_back(this);
```
manishtwari, line 21: noticed that we're updating the static vector `getAllWatchers().push_back(this)` during construction. If multiple watchers are spawned, could this potentially lead to a race on the vector? same might apply for other places where this vector is accessed.

src/manager.cpp
```diff
@@ -46,9 +47,74 @@ sdbusplus::async::task<> Manager::init()
...
+    }
+    return watchedPaths;
+}
+
+void Manager::logSignalHandler(int signal)
+{
+    lg2::info("Received signal: {SIGNAL}", "SIGNAL", signal);
+    std::unordered_set<std::string> watchedPaths = collectAllWatchedPaths();
+
+    std::string checkForWatchedPaths = "/tmp/check_watched_paths.txt";
```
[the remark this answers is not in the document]
spinler: I think keeping at top level /tmp is probably fine, but maybe put something data sync related in the name so it's obvious what it's for, including if it's ever added to a BMC dump.

RameshIyyar:
Its addressed in the https://github.com/ibm-openbmc/phosphor-data-sync/pull/104 PR with different approach so closing it.

PR #34, opened 2025-07-07
rsync : sync only the modified data path via Rsync

aryakpadman:
The commit modifies the syncData API which enables RSYNC to sync only the modified data path if available as SRC path instead of syncing the configured data path every time.

The change also modifies the file paths data type to std::filesystem::path for optimized performance.

Change-Id: Iac4bb9367ab84b34193e577999ae417ab38fe1e5

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -148,7 +148,8 @@ sdbusplus::async::task<> Manager::startSyncEvents()
 // concurrent sync changes.
 sdbusplus::async::task<bool>
     // NOLINTNEXTLINE
-    Manager::syncData(const config::DataSyncConfig& dataSyncCfg)
+    Manager::syncData(const config::DataSyncConfig& dataSyncCfg,
+                      fs::path srcPath)
 {
     using namespace std::string_literals;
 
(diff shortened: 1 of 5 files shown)


src/manager.cpp
```diff
@@ -158,7 +159,15 @@ sdbusplus::async::task<bool>
         "rsync --compress --recursive --perms --group --owner --times --atimes"
         " --update --relative --delete --delete-missing-args "};
 
-    syncCmd.append(" "s + dataSyncCfg._path);
+    if (!srcPath.empty())
+    {
+        // Configure modified path as SRC path rsync if available.
```
RameshIyyar, line 164: confusing the comment...

I think you meant like, use the given source path if available

src/manager.cpp
```diff
@@ -219,9 +228,9 @@ sdbusplus::async::task<>
                 {
                     break;
                 }
-                for ([[maybe_unused]] const auto& dataOp : dataOperations)
+                for (const auto& dataOp : dataOperations)
                 {
-                    co_await syncData(dataSyncCfg);
+                    co_await syncData(dataSyncCfg, std::get<0>(dataOp));
```
RameshIyyar, line 233: Actually, we no need to use `get` API, just use `dataOp.first`.

also, I don't think we need to use `std::map`..could be `std::vector`

PR #35, opened 2025-07-07
config: Add sync config for host console log files

manishtwari:
This commit adds '/var/log/obmc-console.log' and '/var/log/ obmc-console1.log' to the sync configuration to ensure host boot and IPL logs are available on both BMCs

These files capture critical boot-related logs, including SBE and Hostboot (PHYP and OS) istep data. Syncing them helps preserve historical context of host boot attempts, which is useful for debugging, especially after a BMC failover

### DIFF
--- a/config/data_sync_list/common.json
+++ b/config/data_sync_list/common.json
@@ -1,4 +1,20 @@
 {
+    "Files": [
+        {
+            "Path": "/var/log/obmc-console.log",
+            "Description": "Host console persisted log data",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Periodic",
+            "Periodicity": "PT60S"
+        },
+        {
+            "Path": "/var/log/obmc-console1.log",
+            "Description": "Host console1 persisted log data",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Periodic",
+            "Periodicity": "PT60S"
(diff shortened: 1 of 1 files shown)


config/data_sync_list/common.json
```diff
@@ -1,4 +1,20 @@
 {
+    "Files": [
+        {
+            "Path": "/var/log/obmc-console.log",
+            "Description": "Host console persisted log data",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Periodic",
```
RameshIyyar, line 7: Actually, it doesn't need to be periodic since the update only happens during IPL, and such changes are rare in the field. Once we have deferred sync support (or the state-driven sync), it could be updated accordingly. But for now, this approach should be fine.

config/data_sync_list/common.json
```diff
@@ -1,4 +1,20 @@
...
+            "Periodicity": "PT60S"
```
RameshIyyar, line 8: I feel interval is fine but, I would defer to data owner to share suggestion whether it need to decrease or increase.

RameshIyyar (approved):

harsh-agarwal1 (approved):

RameshIyyar:
add to approvelist

PR #44, opened 2025-07-25
Replace `std::system` with `posix_spawn` for async execution

aryakpadman:
This commit introduces the changes to replace the `std::system` with `posix_spawn` for the execution of RSYNC CLI in order to achieve the asynchronous execution of multiple RSYNC commands.

Reasons for moving to posix_spawn():

 - The `std::system()` is a blocking call where it hold until the command finishes execution which is achieving asynchronous behavior.
 - `std::system()` lacks the method to capture the output of the execution to the parent/caller and the return value of the execution which results in limited error handling.
 - `posix_spawn` enables launching child processes without blocking the main thread.
 - `posix_spawn_file_actions` allows fine-grained control of fd redirection, useful for isolating command input/output.

Details of implementation:

 - The API `execCmd()` invokes `posix_spawn()` to execute the rsync command asynchronously.
 - Redirects child's stdout/stderr through a pipe to capture and process output at the parent.
 - Uses `posix_spawn_file_actions_t` to safely setup file descriptors.
 - The API `waitForCmdCompletion()` waits on the child process and reads output from the pipe.
 - The `execCmd()` finally returns the result code and output string to the caller.

Tested:
 - Verified the sync behavior on Huygens simics model during full sync and immediate sync.

Change-Id: I2f80f7806db62a022425df9e84a7edeb28d355e9

### DIFF
--- a/src/async_command_exec.cpp
+++ b/src/async_command_exec.cpp
@@ -1,0 +1,216 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#include "async_command_exec.hpp"
+
+#include <phosphor-logging/lg2.hpp>
+
+namespace data_sync::async
+{
+
+namespace utility
+{
+
+SpawnFActions::SpawnFActions()
+{
+    if (posix_spawn_file_actions_init(&_actions) != 0)
+    {
+        lg2::error("Failed to init posix_spawn_file_actions, errno : {ERRNO}, "
+                   "ERROR : {ERROR}",
+                   "ERRNO", errno, "ERROR", strerror(errno));
+        throw std::runtime_error("Failed to init posix_spawn_file_actions");
+    }
+}
+
+SpawnFActions::~SpawnFActions()
+{
+    if (posix_spawn_file_actions_destroy(&_actions) != 0)
+    {
+        lg2::error(
+            "Failed to destroy the file action instance, errno : {ERRNO}, "
+            "ERROR : {ERROR}",
+            "ERRNO", errno, "ERROR", strerror(errno));
+    }
+}
+
+posix_spawn_file_actions_t* SpawnFActions::get()
+{
+    return &_actions;
+}
+
+} // namespace utility
+
+AsyncCommandExecutor::AsyncCommandExecutor(sdbusplus::async::context& ctx) :
+    _ctx(ctx)
+{}
+
+bool AsyncCommandExecutor::setupPipe(int pipefd[2])
+{
+    if (pipe(pipefd) == -1)
+    {
+        lg2::error("Failed to create pipe. Errno: {ERRNO}, Error: {MSG}",
+                   "ERRNO", errno, "MSG", strerror(errno));
+        return false;
+    }
+    return true;
+}
+
+bool AsyncCommandExecutor::setupPipeRedirection(const FD& readFd,
+                                                const FD& writeFd,
+                                                const auto& actions)
+{
+    // Duplicate the write end of the pipe with the child's stdout and stderr
+    // so that the parent can read child stdout and stderr after the command
+    // execution is completed if any.
+    if (posix_spawn_file_actions_adddup2(actions, writeFd(), STDOUT_FILENO) !=
+            0 ||
+        posix_spawn_file_actions_adddup2(actions, writeFd(), STDERR_FILENO) !=
+            0)
+    {
+        lg2::error(
+            "Failed to duplicate the STDOUT/STDERR to pipe. Errno : {ERRNO}, Error : {MSG}",
+            "ERRNO", errno, "MSG", strerror(errno));
+        return false;
+    }
+
+    // Close the read end of the pipe in child before executing command because
+    // the parent only needs to read.
+    // Otherwise, kernel will think that the child is also reading and will be
+    // open even after parent closes it read end, which could cause blocking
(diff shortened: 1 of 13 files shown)


src/manager.cpp
```diff
@@ -199,6 +224,159 @@ sdbusplus::async::task<bool>
     co_return true;
 }
 
+sdbusplus::async::task<std::pair<int, std::string>>
```
RameshIyyar, line 227: I feel we should move functions into a separate class and make it generic like AsyncCommandExecutor.

Also, please use RAII style to close FDs and Destroy fileaction object instead of manual.

src/async_command_exec.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    int exitCode = WIFEXITED(status) ? WEXITSTATUS(status) : -1;
+    if (!WIFEXITED(status))
+    {
+        lg2::error("Child exited abnormally. Status: {STATUS}", "STATUS",
+                   status);
+    }
+    else if (exitCode != 0)
+    {
+        lg2::error("Command failed with  exitCode: {CODE}, Output: {OUT}",
```
RameshIyyar, line 117: Replace output tag with error in the trace.

src/async_command_exec.cpp
```diff
@@ -0,0 +1,175 @@
...
+        }
+        else
+        {
+            lg2::error("read failed on fd[{FD}] : [{ERROR}]", "FD", fd, "ERROR",
+                       strerror(errno));
+            break;
+        }
+    }
+
+    fdioInstance.reset();
```
RameshIyyar, line 170: Do we need to reset explicitly?

src/manager.cpp
```diff
@@ -2,8 +2,12 @@
 
 #include "manager.hpp"
 
+#include "async_command_exec.hpp"
 #include "data_watcher.hpp"
 
+#include <fcntl.h>
```
RameshIyyar, line 8: These two includes require, here?

src/manager.cpp
```diff
@@ -197,9 +198,10 @@ sdbusplus::async::task<bool>
     // Add destination data path if configured
     syncCmd.append(dataSyncCfg._destPath.value_or(fs::path("")));
     lg2::debug("RSYNC CMD : {CMD}", "CMD", syncCmd);
-    int result = std::system(syncCmd.c_str()); // NOLINT
 
-    if (result != 0)
+    data_sync::async::AsyncCommandExecutor executor(_ctx);
+    auto result = co_await executor.execCmd(syncCmd); // NOLINT
+    if (result.first != 0)
```
RameshIyyar, line 204: Could you please output (which will have error message) in the trace?

src/async_command_exec.cpp
```diff
@@ -0,0 +1,175 @@
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
aryakpadman: Updated

src/async_command_exec.cpp
```diff
@@ -0,0 +1,175 @@
...
+    {
+        lg2::error("Failed to create pipe. Errno : {ERRNO}, Error : {MSG}",
+                   "ERRNO", errno, "MSG", strerror(errno));
+        co_return {-1, ""};
+    }
+
+    FD readFd(pipefd[0]);
+    FD writeFd(pipefd[1]);
+
+    utility::SpawnActions fileActions;
```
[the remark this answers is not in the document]
aryakpadman: Ok changed the class name to PosixSpawnFActions

Defining multiple wrappers as 
1. setupPipe() 
2. setupPipeRedirection()
3. spawnCommand()

considered a single wrapper, but that would require defining writeFD, readFD, etc. inside execCmd and then passing them around as out parameters, or managing them via unique pointers. Both approach felt as bringing complexity. Hence went with separate wrappers.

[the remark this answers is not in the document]
RameshIyyar: We had discussed decoupling the POSIX dependency from this class but anyway its upto you, no functional issues.

src/async_command_exec.cpp
```diff
@@ -0,0 +1,216 @@
+// SPDX-License-Identifier: Apache-2.0
+
```
RameshIyyar, line 2: Commit message: 

> - Introduced another class `SpawnActions` which initializes
   `posix_spawn_file_actions_t` inorder safely setup and destroy file
   descriptors.

Typo in class name or missed to update while the changing the class name.

RameshIyyar (commented):
minor comments.

src/async_command_exec.hpp
```diff
@@ -0,0 +1,155 @@
...
+     *        'posix_spawn'.
+     *
+     * @param[in] - cmd - The bash command to execute
+     *
+     * @return sdbusplus::async::task<std::pair<int, std::string>>
+     *              - int : Exit code of the spawned process (-1 on failure)
+     *              - std::string : Combined stdout and stderr output
+     */
+    sdbusplus::async::task<std::pair<int, std::string>>
+        execCmd(const std::string& cmd);
```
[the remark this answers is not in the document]
aryakpadman: Will `SpawnFActions` make sense?

spinler (approved):

PR #48, opened 2025-07-29
config: Add sync config for encrypted credential store data

manishtwari:
This commit adds '/etc/credstore.encrypted/' to the sync configuration to ensure encrypted credential data is preserved across BMC

The directory is synced immediately from the active to the passive

### DIFF
--- a/config/data_sync_list/common.json
+++ b/config/data_sync_list/common.json
@@ -41,6 +41,12 @@
             "Description": "System LED persisted data",
             "SyncDirection": "Active2Passive",
             "SyncType": "Immediate"
+        },
+        {
+            "Path": "/etc/credstore.encrypted/",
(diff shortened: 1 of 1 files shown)


config/data_sync_list/common.json
```diff
@@ -28,6 +28,12 @@
             "SyncDirection": "Active2Passive",
             "SyncType": "Immediate",
             "ExcludeFilesList": ["/var/lib/phosphor-fan-presence/control/"]
+        },
+        {
+            "Path": "/etc/credstore.encrypted/",
```
RameshIyyar, line 33: Looks good to me but, I would like to confirm whether the credstore.encrypted could be common or IBM specific. @joseph-reynolds

I assume we could sync it i.e its not specific to the BMC...

PR #56, opened 2025-08-08
Include list implementation

aryakpadman:


manishtwari (approved):

src/data_watcher.cpp
```diff
@@ -178,8 +178,83 @@ bool DataWatcher::isPathExcluded(const fs::path& path)
...
+        lg2::debug("{PATH} present inside include list", "PATH",
+                   normalizedPath);
+        return true;
+    }
+
+    // check whether path is child of any includelist path.
+    auto isParentOfNormPath = [&normalizedPath](const auto& includePath) {
+        auto [parentItr, childItr] = std::ranges::mismatch(includePath,
+                                                           normalizedPath);
+        return static_cast<bool>((parentItr == includePath.end()) ||
```
spinler, line 211: the static_cast shouldn't be necessary?

aryakpadman: Sorry I am not able to recall whether I added based on clang-tidy suggestion. Now rerun CI and confirmed that it is not throwing any errors without static-cast. 
`auto` also is not mandating casting as it evaluating to bool on both sides.
Hence removed.

src/data_watcher.cpp
```diff
@@ -178,37 +178,169 @@ bool DataWatcher::isPathExcluded(const fs::path& path)
...
+        // iterator for both paths which will cause ambiguity.
+        std::error_code ec;
+        if (fs::equivalent(normalizedPath, includePath, ec))
+        {
+            return false;
+        }
+
+        auto [parentItr, childItr] = std::ranges::mismatch(normalizedPath,
+                                                           includePath);
+        return static_cast<bool>((parentItr == normalizedPath.end()) ||
```
spinler, line 245: still a static_cast here.

PR #58, opened 2025-08-18
config: Add sync config for system inventory data

manishtwari:
This commit adds '/var/lib/phosphor-inventory-manager/' to the sync configuration to ensure system inventory data is preserved across BMC

The data is synced immediately from the active to the passive BMC to retain inventory state. However, the sync destination on the passive side is set to '/var/lib/phosphor-data-sync/bmc_data_bkp/' to avoid triggering inventory publication on DBus during passive BMC reboot

### DIFF
--- a/config/data_sync_list/common.json
+++ b/config/data_sync_list/common.json
@@ -50,6 +50,13 @@
             "ExcludeList": [
                 "/var/lib/phosphor-settings-manager/settings/xyz/openbmc_project/control/minimum_ship_level_required__"
             ]
+        },
+        {
+            "Path": "/var/lib/phosphor-inventory-manager/",
+            "Description": "System inventory persisted data (synced to backup dir on passive to avoid DBus exposure)",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Immediate",
+            "DestinationPath": "/var/lib/phosphor-data-sync/bmc_data_bkp/"
         }
     ]
 }


config/data_sync_list/common.json
```diff
@@ -41,6 +41,13 @@
             "Description": "System LED persisted data",
             "SyncDirection": "Active2Passive",
             "SyncType": "Immediate"
+        },
+        {
+            "Path": "/var/lib/phosphor-inventory-manager/",
+            "Description": "System inventory persisted data",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Immediate",
+            "DestinationPath": "/var/lib/phosphor-data-sync/bmc_data_bkp/"
```
RameshIyyar, line 50: What happens if this backup directory is not exist? does it fail?

config/data_sync_list/common.json
```diff
@@ -41,6 +41,13 @@
             "Description": "System LED persisted data",
             "SyncDirection": "Active2Passive",
             "SyncType": "Immediate"
+        },
+        {
+            "Path": "/var/lib/phosphor-inventory-manager/",
+            "Description": "System inventory persisted data",
```
RameshIyyar, line 47: Update description, why its synced into backup directory?

RameshIyyar (approved):

config/data_sync_list/common.json
```diff
@@ -50,6 +50,13 @@
             "ExcludeList": [
                 "/var/lib/phosphor-settings-manager/settings/xyz/openbmc_project/control/minimum_ship_level_required__"
             ]
+        },
+        {
+            "Path": "/var/lib/phosphor-inventory-manager/",
+            "Description": "System inventory persisted data (synced to backup dir on passive to avoid DBus exposure)",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Immediate",
+            "DestinationPath": "/var/lib/phosphor-data-sync/bmc_data_bkp/"
```
SunnySrivastava1984, line 59: Is this path dedicated to PIM data or all the back up's will be at this path itself?

manishtwari: This backup path is only for the PIM data
We didn’t want the passive BMC to load and publish inventory on DBus that’s why we’re not syncing it directly to the actual PIM path. Instead, we’re putting it in a backup location to avoid any DBus exposure on the passive side

RameshIyyar: > This backup path is only for the PIM data

Its not correct...

The backup path is for all component but, we will able to get unique path of that the component..

For example,

```
/var/lib/phosphor-data-sync/bmc_data_bkp/var/lib/phosphor-inventory-manager/
/var/lib/phosphor-data-sync/bmc_data_bkp/var/lib/vpd
...
/var/lib/phosphor-data-sync/bmc_data_bkp/...
```

SunnySrivastava1984: ok, That is what I was looking for.

PR #61, opened 2025-08-19
config: Add sync config for settings manager data

manishtwari:
This commit adds '/var/lib/phosphor-settings-manager/settings/' to the sync configuration to preserve system settings across BMC

The data is synced immediately from active to passive BMC. The file 'minimum_ship_level_required__' is excluded from sync

### DIFF
--- a/config/data_sync_list/common.json
+++ b/config/data_sync_list/common.json
@@ -41,6 +41,15 @@
             "Description": "System LED persisted data",
             "SyncDirection": "Active2Passive",
             "SyncType": "Immediate"
+        },
+        {
+            "Path": "/var/lib/phosphor-settings-manager/settings/",
+            "Description": "Persisted settings data",
+            "SyncDirection": "Active2Passive",
+            "SyncType": "Immediate",
+            "ExcludeList": [
+                "/var/lib/phosphor-settings-manager/settings/xyz/openbmc_project/control/minimum_ship_level_required__"
+            ]
         }
     ]
 }


config/data_sync_list/common.json
```diff
@@ -41,6 +41,15 @@
             "Description": "System LED persisted data",
             "SyncDirection": "Active2Passive",
             "SyncType": "Immediate"
+        },
+        {
+            "Path": "/var/lib/phosphor-settings-manager/settings/",
```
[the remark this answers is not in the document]
RameshIyyar: Yes, that’s right. I think we can update it once the sync notification support is in place. @aryakpadman fyi.

config/data_sync_list/common.json
```diff
@@ -41,6 +41,15 @@
...
+            "Description": "Persisted settings data",
+            "SyncDirection": "Active2Passive",
```
RameshIyyar, line 48: Is it Active to Passive? Since the settings manager runs on both BMCs, I assume some settings could also be updated by the Passive BMC. The settings manager won't controls which settings are allowed on Active vs Passive BMC. Ultimately, this will be managed through the Redfish client (ASMI/HMC), so I think it needs to be bidirectional—but I’d defer to @geissonator.

geissonator: I thought when we last discussed, we said we would not need to run the settings-manager on the passive BMC (only start during Active). But I also know we've flip flopped a lot in discussions. I'm not seeing anything obvious, that would be needed, other then maybe the time setting:

```
root@p11bmc:~# ls -al /var/lib/phosphor-settings-manager/settings/xyz/openbmc_project/control/host0/
drwxr-xr-x    3 root     root          4096 Aug 26 14:17 .
drwxr-xr-x    3 root     root          4096 Jul 10 07:45 ..
-rw-r--r--    1 root     root            86 Sep  2 14:23 TPMEnable__
-rw-r--r--    1 root     root            88 Sep  2 14:23 auto_reboot__
-rw-r--r--    1 root     root           133 Sep  2 14:23 boot__
-rw-r--r--    1 root     root           118 Sep  2 14:23 power_cap__
drwxr-xr-x    2 root     root          4096 Aug  7 14:50 power_restore_policy
root@p11bmc:~# ls -al /var/lib/phosphor-settings-manager/settings/xyz/openbmc_project/
drwxr-xr-x    5 root     root          4096 Jul 24 12:03 .
drwxr-xr-x    3 root     root          4096 Jul 10 07:45 ..
drwxr-xr-x    3 root     root          4096 Jul 10 07:45 control
drwxr-xr-x    2 root     root          4096 Jul 17 06:09 software
drwxr-xr-x    2 root     root          4096 Jul 24 12:03 time
root@p11bmc:~# ls -al /var/lib/phosphor-settings-manager/settings/xyz/openbmc_project/software/
drwxr-xr-x    2 root     root          4096 Jul 17 06:09 .
drwxr-xr-x    5 root     root          4096 Jul 24 12:03 ..
-rw-r--r--    1 root     root            92 Aug 25 05:25 apply_time__
```

I think for now this is good and if we have to revisit once we start bringing things up we can always come back and change things here.

PR #62, opened 2025-09-01
Sibling notification on Local BMC

aryakpadman:


src/data_sync_config.hpp
```diff
@@ -71,6 +71,37 @@ struct Retry
     std::chrono::seconds _retryIntervalInSec;
 };
 
+/**
+ * @brief Configuration for notifying the sibling BMC after a successful sync.
+ *
+ * Stores information on which paths to notify and how to notify the sibling
+ * BMC.
+ */
+using notifyOnPaths = std::unordered_set<fs::path>;
```
manishtwari, line 80: nit: it should be good to have alias name in pascal case `NotifyOnPaths`

RameshIyyar (commented):
I feel this PR needs reworking with the POSIX and INC PRs, as there are some dependencies. I haven’t completed my review of the last two commits yet, so I’m posting comments based on what I’ve found so far.

config/schema/schema.json
```diff
@@ -144,6 +150,40 @@
...
+                    "$ref": "#/$defs/notifyServices"
+                }
+            },
+            "required": ["Mode", "NotifyServices"]
+        },
+        "notifySiblingForDirs": {
+            "description": "The JSON object which describes how the data owner to be notified once the data got changed",
+            "type": "object",
+            "properties": {
+                "NotifyOnPaths": {
```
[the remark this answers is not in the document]
aryakpadman: Felt `PathsToNotify`  sounds like the targets we're sending the notification, whereas the intention is our intention is to convey 'when these paths change, trigger notification'.  Doesn’t `NotifyOnPaths` express that?

aryakpadman:
This PR is dependent on includeList PR changes and hence has those commits also. Hence please review only the last 5 commits.

src/manager.cpp
```diff
@@ -233,6 +283,46 @@ sdbusplus::async::task<bool>
 
         co_return false;
     }
+    else if (dataSyncCfg._notifySibling.has_value())
+    {
+        if (dataSyncCfg._notifySibling.value()._paths.has_value())
```
RameshIyyar, line 293: Move all this logic into triggerSiblingNotification, since that function is responsible for triggering the notification. The decision of whether a notification is required and whether it succeeds should both be handled within that function.

src/manager.cpp
```diff
@@ -164,55 +167,102 @@ sdbusplus::async::task<> Manager::startSyncEvents()
...
-        " --update --relative --delete --delete-missing-args "};
+    if (mode == RsyncMode::Sync)
+    {
+        // Appending required flags to sync data between BMCs
+        // For more details about CLI options, refer rsync man page.
+        // https://download.samba.org/pub/rsync/rsync.1#OPTION_SUMMARY
 
-    if (dataSyncCfg._excludeList.has_value())
+        cmd.append(
+            "rsync --compress --recursive --perms --group --owner --times --atimes"
```
RameshIyyar, line 183: All these common option could be in default and append necessary option `--remove-source-files` if the mode is notify to avoid updating command in two different places.

src/manager.cpp
```diff
@@ -283,8 +283,13 @@ sdbusplus::async::task<bool>
 
         co_return false;
     }
-    else if (dataSyncCfg._notifySibling.has_value())
+    else if ((dataSyncCfg._notifySibling.has_value()) && (result.first == 0) &&
```
RameshIyyar, line 286: > (result.first == 0)

why its required? it will come only if the result.first is zero.

PR #63, opened 2025-09-08
config: Add sync config for service manager data

manishtwari:
Added '/var/lib/service-config-manager/' to the sync list to make sure service enable/disable state is preserved across BMC
The data is synced immediately from active to passive BMC

### DIFF
--- a/config/data_sync_list/common.json
+++ b/config/data_sync_list/common.json
@@ -57,6 +57,12 @@
             "SyncDirection": "Active2Passive",
             "SyncType": "Immediate",
             "DestinationPath": "/var/lib/phosphor-data-sync/bmc_data_bkp/"
+        },
+        {
+            "Path": "/var/lib/service-config-manager/",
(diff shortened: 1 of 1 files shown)


config/data_sync_list/common.json
```diff
@@ -57,6 +57,12 @@
             "SyncDirection": "Active2Passive",
             "SyncType": "Immediate",
             "DestinationPath": "/var/lib/phosphor-data-sync/bmc_data_bkp/"
+        },
+        {
+            "Path": "/var/lib/service-config-manager/",
+            "Description": "Persisted service enable/disable config data",
+            "SyncDirection": "Active2Passive",
```
RameshIyyar, line 64: We recently discussed that the service manager needs to run on both BMCs, so this should also be treated as bidirectional with notifications. We just need to ensure that the file path remains unique if both BMCs can update it, to avoid race conditions.

@geissonator any comments?

PR #64, opened 2025-10-06
Add retry logic for sync failures

manishtwari:


meson.options
```diff
@@ -15,12 +15,12 @@ option(
 # failure unless overridden from respective JSON file configuration.
 # Default value will be 3.
 # A retry attempt value of zero indicates no retries will be performed.
-option('retry_attempts', type: 'integer', min: 0, value: 3)
+option('retry_attempts', type: 'integer', min: 0, value: 1)
```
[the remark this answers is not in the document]
manishtwari: Yep, updated retry is now 2, so total 3 sync attempts including the first

meson.options
```diff
@@ -15,12 +15,12 @@ option(
...
 
 # The retry interval in seconds which is applicable for all files/directories
 # during sync retry unless overridden from respective JSON file configuration.
 # Default value is 5secs.
-option('retry_interval', type: 'integer', value: 5)
+option('retry_interval', type: 'integer', value: 20)
```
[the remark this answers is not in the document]
manishtwari: Yeah, I had made it 20s earlier since one of the full sync tests was failing due to timeout, but now changed it back to 30s and using smaller interval in test config

src/manager.cpp
```diff
@@ -213,27 +220,52 @@ sdbusplus::async::task<bool>
...
+                "OLD", currentSrcPath, "NEW", vanishedPath);
 
-        co_return false;
+            co_await sleep_for(_ctx, std::chrono::seconds(retryIntervalSec));
+            co_return co_await syncData(dataSyncCfg, vanishedPath,
+                                        retryCount + 1);
+        }
+        else
+        {
+            lg2::info(
```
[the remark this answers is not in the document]
manishtwari: not really needed at info, I’ll move it to debug since it’s an expected case, not something we need in normal logs

src/manager.cpp
```diff
@@ -170,6 +170,11 @@ sdbusplus::async::task<bool>
                       fs::path srcPath, std::vector<fs::path> vanishedPaths,
                       size_t retryCount)
 {
+    if (_syncBMCDataIface.disable_sync())
```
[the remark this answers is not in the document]
manishtwari: I think it’s better to keep this check inside syncData() instead of the retry function. This way, if sync is disabled, we just return early without going into retry logic
yeah, I’ll remove the duplicate checks from immediate and periodic sync since syncData() already handles it now

manishtwari (commented):
pushed the updated changes addressing all review comments. 
Let me know if anything still looks off.

src/manager.cpp
```diff
@@ -262,34 +263,57 @@ sdbusplus::async::task<void>
...
-            "PATH", srcPath, "CFGPATH", dataSyncCfg._path, "ERRCODE",
-            result.first, "ERROR", result.second);
+        lg2::info(
+            "Sibling notification succeeded for [{SRC}] under [{CFGPATH}]",
+            "SRC", srcPath, "CFGPATH", dataSyncCfg._path);
+        co_return;
     }
-    else
+
+    if (retryCount++ < dataSyncCfg._retry->_retryAttempts)
```
manishtwari: I kept the retry logic inside `triggerSiblingNotification()` itself, just to keep it simple
I could’ve used the `retrySync()` API, but then I’d have to pass the mode and add extra if-else checks for both sync and notify, which didn’t really make sense
This way it’s cleaner and easier to follow, but I’m open to suggestions if you think there’s a better way

src/manager.cpp
```diff
@@ -355,7 +355,9 @@ sdbusplus::async::task<>
                 }
                 for (const auto& [path, dataOp] : dataOperations)
                 {
-                    co_await syncData(dataSyncCfg, path);
+                    _ctx.spawn(
+                        syncData(dataSyncCfg, path) |
+                        stdexec::then([]([[maybe_unused]] bool result) {}));
```
RameshIyyar: why is it required? 

are you getting any error?

src/manager.cpp
```diff
@@ -277,51 +290,99 @@ sdbusplus::async::task<void>
...
 {
+    constexpr size_t rsyncVanishedExitCode = 24;
+    const fs::path currentSrcPath = srcPath.empty() ? dataSyncCfg._path
+                                                    : srcPath;
     std::string syncCmd{};
-    getRsyncCmd(RsyncMode::Sync, dataSyncCfg, srcPath.string(), syncCmd);
+    const bool haveIncList = getRsyncCmd(RsyncMode::Sync, dataSyncCfg,
+                                         srcPath.string(), syncCmd);
+
+    if (haveIncList)
```
RameshIyyar: I’m not following this — why is it needed?

I don’t see any explanation in the commit message, and the variable name itself is a bit confusing.

src/manager.cpp
```diff
@@ -277,51 +290,99 @@ sdbusplus::async::task<void>
...
+    co_return false;
+}
+
 sdbusplus::async::task<bool>
     // NOLINTNEXTLINE
     Manager::syncData(const config::DataSyncConfig& dataSyncCfg,
-                      fs::path srcPath)
+                      fs::path srcPath, size_t retryCount)
 {
+    constexpr size_t rsyncVanishedExitCode = 24;
```
RameshIyyar: > rsyncVanishedExitCode

RsyncFilesVanishedExitCode - it should be UpperCamelCase

src/manager.hpp
```diff
@@ -188,8 +188,12 @@ class Manager
      * @param[in] srcPath - The modified path inside the cfg path.
      *                      Will be empty if not available.
      * @param[out] cmd - string where the framed RSYNC command holds.
+     *
+     * @return true if include-list is present in the configuration, false
```
RameshIyyar: I believe the intention is to determine whether syncing is required based on the existence of the included paths, correct?

If that’s the case, you could rely on the constructed command arguments to make that decision — for example, if the command string ends up empty, simply avoid running the sync. Currently, the logic continues to construct the rsync CLI even after deciding that syncing isn’t needed.

Please update the function’s Doxygen comments to document this behavior.

I would suggest add details about this in the commit message or push as separate commit.

src/manager.hpp
```diff
@@ -200,13 +204,28 @@ class Manager
...
-                 fs::path srcPath = fs::path{});
+                 fs::path srcPath = fs::path{}, size_t retryCount = 0);
+
+    /**
+     * @brief Retry the data sync operation based on failure
+     *
+     * @param[in] cfg - Data sync configuration
+     * @param[in] srcPath - Source path to be synced
+     * @param[in] retryCount - Current retry attempt number
+     * @param[in] rsyncExitCode - Exit code from the last rsync attempt
```
RameshIyyar: why is it required? is it really worth? 

We will have error trace the sync is failed, right? 

Please avoid adding redundant traces.

src/manager.cpp
```diff
@@ -202,10 +203,21 @@ void Manager::getRsyncCmd(RsyncMode mode,
...
-        std::ranges::for_each(dataSyncCfg._includeList.value(), appendToCmd);
+
+        if (std::ranges::none_of(dataSyncCfg._includeList.value(), exists_now))
+        {
+            lg2::info(
+                "IncludeList: none of the configured source paths exist, skipping rsync");
+            haveIncList = true;
+        }
+
+        std::ranges::for_each(
```
RameshIyyar: I think you are checking existence twice so it could tweaked as like below
```
// Paths that do not yet exist will be synced once they are created; otherwise, the sync will fail.

auto existPath = [](const fs::path& p) {
    std::error_code ec;
    return fs::exists(p, ec);
};

bool anyExist = false;
std::ranges::for_each(dataSyncCfg._includeList.value() | std::views::filter(existPath),
    [&](const fs::path& p) {
        cmd.append(" "s + p.string());
        anyExist = true;
    });

if (!anyExist)
{
    lg2::debug("None of the configured include paths are exist, skipping rsync");
    cmd.clear();
    return
}
```

src/manager.cpp
```diff
@@ -277,51 +290,99 @@ sdbusplus::async::task<void>
...
+    const fs::path currentSrcPath = srcPath.empty() ? dataSyncCfg._path
+                                                    : srcPath;
     std::string syncCmd{};
-    getRsyncCmd(RsyncMode::Sync, dataSyncCfg, srcPath.string(), syncCmd);
+    const bool haveIncList = getRsyncCmd(RsyncMode::Sync, dataSyncCfg,
+                                         srcPath.string(), syncCmd);
+
+    if (haveIncList)
+    {
+        // IncludeList with no existing entries no-op success
```
RameshIyyar: It seems sync is skipped if none of the configured include paths exist, otherwise it fails during the full sync.

But, what happens if a configured path does not exist? Should we handle that case as well?

src/manager.cpp
```diff
@@ -334,6 +337,19 @@ sdbusplus::async::task<bool>
     constexpr size_t rsyncVanishedExitCode = 24;
     const fs::path currentSrcPath = srcPath.empty() ? dataSyncCfg._path
                                                     : srcPath;
+
+    if (retryCount == 0)
+    {
+        if (dataSyncCfg._syncInProgressPaths.find(currentSrcPath) !=
```
RameshIyyar: You could use `contains` API instead of `find` and `end` APIs and these two `if` block could be combined 

```
if (retryCount == 0 && dataSyncCfg._syncInProgressPaths.contains(currentSrcPath))
{
    lg2::debug("Skipping sync for [{SRC}] since its already in progress", "SRC", currentSrcPath);
    co_return false;
}
```

src/manager.cpp
```diff
@@ -334,6 +337,19 @@ sdbusplus::async::task<bool>
...
+            dataSyncCfg._syncInProgressPaths.end())
+        {
+            lg2::info("Skipping sync for [{SRC}]: already in progress", "SRC",
```
RameshIyyar: Don't use info trace, it will be more.

src/manager.cpp
RameshIyyar: I think you are using the `_syncInProgressPaths` container to hold syncing paths - what happens for the include path? did you test this case?

src/manager.cpp
```diff
@@ -202,10 +203,21 @@ void Manager::getRsyncCmd(RsyncMode mode,
...
-            cmd.append(" "s + path.string());
+        auto exists_now = [](const fs::path& p) {
+            std::error_code ec;
+            return fs::exists(p, ec);
         };
-        std::ranges::for_each(dataSyncCfg._includeList.value(), appendToCmd);
+
+        if (std::ranges::none_of(dataSyncCfg._includeList.value(), exists_now))
+        {
+            lg2::info(
```
aryakpadman: do we need to keep it info? majority of these type of checking logs is in debug level now

src/manager.cpp
```diff
@@ -230,6 +242,7 @@ void Manager::getRsyncCmd(RsyncMode mode,
     {
         cmd.append(" "s + NOTIFY_SERVICES_DIR);
     }
+    return haveIncList;
```
aryakpadman: The API getRsyncCmd was supposed to frame the rsync CLI command. I feel returning true is not aligning to the purpose.

Can we do this check before calling getRsyncCmd() and invoke getRsyncCmd() if includeList paths exists? Will that work?

src/utility.hpp
```diff
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #pragma once
-
```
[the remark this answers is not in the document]
manishtwari: this by mistake, i will revert this

src/manager.cpp
```diff
@@ -319,6 +319,9 @@ sdbusplus::async::task<bool>
     // implemented.
     // setSyncEventsHealth(SyncEventsHealth::Critical);
 
+    // Remove from in-progress after completing all sync attempts
```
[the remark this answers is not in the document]
manishtwari: now using `scope_exit`, it ensures proper cleanup in one place and avoids multiple manual cleanup paths

src/manager.cpp
```diff
@@ -365,7 +381,8 @@ sdbusplus::async::task<bool>
             // initiate sibling notification
             co_await triggerSiblingNotification(dataSyncCfg, srcPath.string());
         }
-
+        // Remove from in-progress after completing sync successfully
+        dataSyncCfg._syncInProgressPaths.erase(currentSrcPath);
```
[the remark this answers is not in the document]
manishtwari: yeah, agreed if we miss clearing the path during an exception any failure, the sync could stall until restart
I checked `std::scope_exit`, it’s ideal for this, but it’s not yet supported in the GCC version we use in OpenBMC, even the latest GCC 15.2 doesn’t have full support
I’ve used `std::experimental::scope_exit` instead, since compiler support is available for it

src/manager.cpp
manishtwari: yeah, I tested the include-list case
at startup, I don’t sync the base `cfg._path`. I expand the IncludeList, take only the paths that actually exist, and build the rsync cmd from those. for tracking, I keep the base path in `_syncInProgressPaths` and clear it once the sync is done
for background sync, if an include path doesn’t exist yet, we watch the parent dir. when that path appears or a file change happens, inotify triggers and we sync that exact path. here the tracking key is the include path itself, so if another inotify event comes, it’s ignored

src/manager.cpp
```diff
@@ -277,51 +277,111 @@ sdbusplus::async::task<void>
...
+                "CODE", result.first, "SRC",
+                srcPath.empty() ? dataSyncCfg._path : srcPath);
+            co_return true;
+        }
 
-        lg2::error(
-            "Error syncing [{PATH}], ErrCode : {ERRCODE}, Error : {ERROR}",
-            "PATH", dataSyncCfg._path, "ERRCODE", result.first, "ERROR",
-            result.second);
+        // Permanent errors — do not retry
```
RameshIyyar: I think 5 also could be considered as permanent error, right?

src/manager.cpp
```diff
@@ -277,51 +277,111 @@ sdbusplus::async::task<void>
...
+        case 1:  // syntax or usage
+        case 2:  // protocol incompatibility
+        case 3:  // input/output paths selection error
+        case 4:  // requested action not supported
+        case 6:  // daemon unable to append to log-file
+        case 11: // error in file I/O
+        case 13: // program diagnostics errors
+        case 14: // Error in IPC code
+        case 22: // Error allocating core memory buffers
```
RameshIyyar: error code 22 is really permanent error?

src/manager.cpp
```diff
@@ -341,6 +342,33 @@ sdbusplus::async::task<bool>
...
                       fs::path srcPath, size_t retryCount)
 {
+    using std::experimental::scope_exit;
+    const fs::path currentSrcPath = srcPath.empty() ? dataSyncCfg._path
+                                                    : srcPath;
+
+    // true only for the first call in the recursive retry chain
+    // marks the top-level attempt that owns the in-progress entry
+    // and will clear it from _syncInProgressPaths after all retries finish
+    bool ownsInProgressCleanup = false;
```
RameshIyyar: I think this flag is not required but no functional issues so its upto you.

src/manager.cpp
manishtwari: For error code 5, it happens when rsync can’t start the client-server handshake mostly seen when daemon/stunnel not ready or some network hiccup, so I kept it retryable for now.
Error code 22 comes when memory allocation fails (OOM), so added it under permanent.
We can update this list later if needed

src/manager.cpp
manishtwari: kept the flag just to make sure we erase only from the scope that actually inserted it (top-level recursion) helps avoid accidental cleanup if any nested retry happens.

RameshIyyar (approved):

src/manager.cpp
```diff
@@ -277,51 +297,148 @@ sdbusplus::async::task<void>
...
+    if (_syncBMCDataIface.disable_sync())
+    {
+        co_return false;
+    }
+
+    using std::experimental::scope_exit;
+    const fs::path currentSrcPath = srcPath.empty() ? dataSyncCfg._path
+                                                    : srcPath;
+
+    auto cleanup = scope_exit([&]() noexcept {
```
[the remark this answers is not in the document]
manishtwari: I agree, it’s generally not ideal to capture everything by reference. In this case, I only needed `dataSyncCfg` and `currentSrcPath`, so I used [&] since both are used inside. I can update it to explicitly capture just those two variables if that’s preferred.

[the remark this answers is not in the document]
spinler: yea.

PR #98, opened 2026-04-23
Fix stale watch descriptors and parent watch handling

aryakpadman:
Issue observed:
Sometimes deletion of the configured path does not add a watch for the parent.

Root cause analysis:
When the configured path is created, a watch is added for it and the parent watch is removed from _watchDescriptors.  Once removed, the kernel emits an IN_IGNORED event for that watch descriptor.

During event logging, the code accesses
_watchDescriptors[receivedEvent->wd], which unintentionally recreates the removed watch descriptor with an empty path, leading to stale entries in the map that are never cleaned up.

These stale entriesinflate _watchDescriptors.size(), causing the size-based check in delete handling to fail. As a result, the parent watch is not re-added when the configured path is deleted.

Fix Proposed:
- Avoid stale entries in _watchDescriptors by using find() instead of [] during event handling, preventing creation of empty entries for IN_IGNORED descriptors.

- Replace unreliable _watchDescriptors.size() check in processDeleteSelf() with a direct check to determine if the deleted path is the configured path or its parent.

Note : This commit updated a logging trace also to print the rsync src path along with rsync stat output for better readability.

Tested:
Verified in rbmc-prototype card that watches are getting added to parent paths as expected and no stale entries in the watchDescriptors map.

Change-Id: I12d41d269555a6befe8449f1e047567be3642ce8

### DIFF
--- a/src/data_watcher.cpp
+++ b/src/data_watcher.cpp
@@ -423,10 +423,17 @@ std::optional<std::vector<EventInfo>> DataWatcher::readEvents()
         // NOLINTNEXTLINE to avoid cppcoreguidelines-pro-type-reinterpret-cast
         auto* receivedEvent = reinterpret_cast<inotify_event*>(&buffer[offset]);
 
+        // Using find() because,
+        // IN_IGNORED events can arrive for already removed watch descriptors
+        // Hence use of [] to access the map entries will cause stale entries.
+        auto wdIter = _watchDescriptors.find(receivedEvent->wd);
+        std::string pathStr = (wdIter != _watchDescriptors.end())
+                                  ? wdIter->second.string()
+                                  : "[removed path]";
+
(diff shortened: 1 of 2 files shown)


src/data_watcher.cpp
```diff
@@ -423,10 +423,17 @@ std::optional<std::vector<EventInfo>> DataWatcher::readEvents()
         // NOLINTNEXTLINE to avoid cppcoreguidelines-pro-type-reinterpret-cast
         auto* receivedEvent = reinterpret_cast<inotify_event*>(&buffer[offset]);
 
+        // Using find() because,
```
riya461, line 426: > These stale entriesinflate _watchDescriptors.size(), causing the

entries inflate ?

RameshIyyar (approved):

riya461 (approved):

PR #101, opened 2026-05-20
datasynctool for ease of user interaction with phosphor-data-sync

aryakpadman:


src/datasynctool/status.cpp
```diff
@@ -0,0 +1,99 @@
...
+    method.append(interface);
+
+    auto reply = bus.call(method);
+    PropertyMap properties;
+    reply.read(properties);
+
+    return properties;
+}
+
+std::string extractEnumValue(const std::string& fullPath)
```
aryakpadman: planning to move extractEnumValue() to utils.hpp

src/datasynctool/utils.cpp
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#include "utils.hpp"
+
+#include <print>
+
+namespace datasynctool::utils
+{
+
+std::string extractEnumValue(const std::string& fullPath)
```
spinler: nit: fullPath to me implies a filesystem or dbus path.

src/datasynctool/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <CLI/CLI.hpp>
+
+#include <iostream>
+
+int main(int argc, char* argv[])
+{
+    CLI::App app{
+        "Data Sync Tool - Command line utility for phosphor-data-sync"};
+
+    // Add status flag
```
spinler: small nit:  The code is pretty clear what it's doing so comments probably not necessary.

src/datasynctool/meson.build
```diff
@@ -0,0 +1,36 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# CLI11 dependency for datasynctool
+cli11_dep = dependency('CLI11', required: false)
+if not cli11_dep.found()
+    cli11_dep = dependency('cli11', required: false)
+endif
+
+# Build datasynctool executable
+if cli11_dep.found()
```
spinler: Is there a reason this is optional?

src/datasynctool/sync_properties.cpp
```diff
@@ -0,0 +1,128 @@
...
+                             const std::string& interface)
+{
+    auto method = bus.new_method_call(service.c_str(), path.c_str(),
+                                      "org.freedesktop.DBus.Properties",
+                                      "GetAll");
+    method.append(interface);
+
+    auto reply = bus.call(method);
+    PropertyMap properties;
+    reply.read(properties);
```
spinler: or:

```
return reply.unpack<propertyMap>();
```

src/datasynctool/dbus_interactions.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+    try
+    {
+        auto bus = sdbusplus::bus::new_default();
+        const std::string service = SyncBMCData::interface;
+        const std::string path = SyncBMCData::instance_path;
+        const std::string interface = SyncBMCData::interface;
+
+        // Log to journal that datasynctool is trying to enable/disable sync
+        lg2::info("datasynctool trying to {ACTION} sync", "ACTION",
```
spinler, line 127: 'trying' implies it may not work.  Is there a reason it wouldn't?

src/datasynctool/dbus_interactions.cpp
```diff
@@ -0,0 +1,128 @@
...
+                  enable ? "enable" : "disable");
+
+        auto method = bus.new_method_call(service.c_str(), path.c_str(),
+                                          "org.freedesktop.DBus.Properties",
+                                          "Set");
+        method.append(interface, "DisableSync", std::variant<bool>(!enable));
+
+        auto reply = bus.call(method);
+
+        std::println("Sync {} successfully", enable ? "enabled" : "disabled");
```
spinler, line 135: same comment as above, if there's no reason to expect the action to fail, could leave this out.

src/datasynctool/sync_properties.cpp
```diff
@@ -0,0 +1,128 @@
...
+        method.append(interface, "DisableSync", std::variant<bool>(!enable));
+
+        auto reply = bus.call(method);
+
+        std::println("Sync {} successfully", enable ? "enabled" : "disabled");
+        return 0;
+    }
+    catch (const sdbusplus::exception_t& e)
+    {
+        std::cerr << "Error setting sync state: " << e.what() << "\n";
```
spinler: Maybe use the journal for these as well so they are captured?

src/datasynctool/fullsync.cpp
```diff
@@ -0,0 +1,49 @@
...
+    try
+    {
+        auto bus = sdbusplus::bus::new_default();
+        const std::string service = SyncBMCData::interface;
+        const std::string path = SyncBMCData::instance_path;
+        const std::string interface = SyncBMCData::interface;
+
+        lg2::info("datasynctool attempting a full sync.");
+
+        auto method = bus.new_method_call(service.c_str(), path.c_str(),
```
spinler: can you not just pass in the constants directly?

src/datasynctool/fullsync.cpp
```diff
@@ -0,0 +1,49 @@
...
+int startFullSync()
+{
+    try
+    {
+        auto bus = sdbusplus::bus::new_default();
+        const std::string service = SyncBMCData::interface;
+        const std::string path = SyncBMCData::instance_path;
+        const std::string interface = SyncBMCData::interface;
+
+        lg2::info("datasynctool attempting a full sync.");
```
spinler: "attempting" sounds like it may or may not work.  How about just 'starting'?

spinler:
Two other random ideas, in case they sound interesting:
1. Add a '-w' flag to the full sync option to wait for it to complete, like obmcutil can on a poweron
2. Also display the PeerConnected property value, since theoretically that has to be Connected for a sync to work.

src/datasynctool/main.cpp
```diff
@@ -0,0 +1,27 @@
...
+    try
+    {
+        app.parse(argc, argv);
+    }
+    catch (const CLI::ParseError& e)
+    {
+        return app.exit(e);
+    }
+
+    // Default behavior when no options are provided
```
RameshIyyar: I think its not required.  You could use `app.set_help_flag`

src/datasynctool/meson.build
RameshIyyar: +1

src/datasynctool/main.cpp
RameshIyyar: +1

src/datasynctool/fullsync.hpp
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+namespace datasynctool::fullsync
+{
+
+/**
+ * @brief Start a full synchronization
+ * @return 0 on success, 1 on error
```
manishtwari: I guess typo -1 on error

src/datasynctool/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        app.parse(argc, argv);
+    }
+    catch (const CLI::ParseError& e)
+    {
+        return app.exit(e);
+    }
+
+    // Handle status option
+    if (showStatus)
```
manishtwari: If user could run this what should happen?
`datasynctool --status --fullSync `
currently only the first matching condition executes due to the early returns, while the remaining options are silently ignored this may confuse user

src/datasynctool/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    // Handle disableSync option
+    if (disableSync)
+    {
+        return datasynctool::sync_properties::setSyncEnabled(false);
+    }
+
+    // Default behavior when no options are provided
+    std::cout << "Data Sync Tool initialized\n";
```
manishtwari: nit: could use `std::println` here as well

src/datasynctool/meson.build
aryakpadman: yeah, It is not mandatory, hence removed.

src/datasynctool/utils.cpp
aryakpadman: Changed to dbusValue.

src/datasynctool/fullsync.cpp
aryakpadman: Used “attempting” because datasynctool just issues the request. The request can still be rejected by the data sync service in cases such as sync or redundancy being disabled.
In that sense, the tool is only issuing the request, not confirming that the sync has started successfully.

Is it fine then?

src/datasynctool/dbus_interactions.cpp
aryakpadman: Thinking a case to restrict enabling sync while redundancy is disabled.

src/datasynctool/dbus_interactions.cpp
aryakpadman: https://github.com/ibm-openbmc/phosphor-data-sync/pull/101/changes/d20dabc89213513e7347273fbeb9dc8dfb91e15b#r3298448713

src/datasynctool/main.cpp
aryakpadman: yeah correct, I made changes so that user can run options together and added group for options to set mutually exclusive options.

Thanks for pointing out.

aryakpadman:
@spinler @manishtwari Addressed the comments, while pushing the previous files `sync_properties.cpp` and `fullsync.cpp` is merged into a single file as `dbus_interactions.cpp`. Please consider this too while reviewing

aryakpadman:
> Two other random ideas, in case they sound interesting:
> 
>     1. Add a '-w' flag to the full sync option to wait for it to complete, like obmcutil can on a poweron
> 
>     2. Also display the PeerConnected property value, since theoretically that has to be Connected for a sync to work.

Yes I will consider . Thanks for the suggestion.

src/datasynctool/fullsync.cpp
spinler: Yea if you'd prefer to keep it that's fine with me.

spinler (approved):

src/datasynctool/main.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include <CLI/CLI.hpp>
+
+#include <print>
+
+int main(int argc, char* argv[])
+{
+    CLI::App app{
+        "Data Sync Tool - Command line utility for phosphor-data-sync"};
+
+    // Parse command line arguments
```
RameshIyyar, line 43: Its not required, the code itself explains it..Please remove unwanted comments.

src/datasynctool/main.cpp
```diff
@@ -0,0 +1,26 @@
...
+    try
+    {
+        app.parse(argc, argv);
+    }
+    catch (const CLI::ParseError& e)
+    {
+        return app.exit(e);
+    }
+
+    std::println("Data Sync Tool initialized");
```
RameshIyyar: Why these two traces are required.

You could use CLI helper usage API, right?

src/datasynctool/dbus_interactions.cpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    return output;
+}
+
+int displayStatus(bool jsonOutput)
+{
+    try
+    {
+        auto bus = sdbusplus::bus::new_default();
```
RameshIyyar: Could you share why you're not preferring the newer asynchronous style of D-Bus interactions?

src/datasynctool/dbus_interactions.cpp
```diff
@@ -0,0 +1,87 @@
...
+        const std::string service = SyncBMCData::interface;
```
RameshIyyar: All of this is unnecessary—you could pass it directly. I understand it was generated by Bob, but please keep an eye on this and avoid such cases going forward.

src/datasynctool/dbus_interactions.cpp
```diff
@@ -0,0 +1,87 @@
...
+{
+
+using SyncBMCData =
+    sdbusplus::common::xyz::openbmc_project::control::SyncBMCData;
+
+PropertyMap getAllProperties(sdbusplus::bus_t& bus, const std::string& service,
+                             const std::string& path,
+                             const std::string& interface)
+{
+    auto method = bus.new_method_call(service.c_str(), path.c_str(),
```
RameshIyyar: Could you share why you're not preferring the newer asynchronous style of D-Bus interactions?

src/datasynctool/main.cpp
```diff
@@ -9,6 +11,13 @@ int main(int argc, char* argv[])
     CLI::App app{
         "Data Sync Tool - Command line utility for phosphor-data-sync"};
 
+    bool showStatus{false};
+    app.add_flag("-s,--status", showStatus,
+                 "Display all D-Bus properties hosted by data sync");
```
RameshIyyar: I don't think we should limit ourselves to displaying only the data sync properties. The PeerConnected property would be a good addition as well, since it is hosted by the Pair application. Even we could add some of the rbmc manager properties too.

src/datasynctool/dbus_interactions.cpp
```diff
@@ -28,6 +32,39 @@ PropertyMap getAllProperties(sdbusplus::bus_t& bus, const std::string& service,
...
+int startFullSync()
+{
+    try
+    {
+        auto bus = sdbusplus::bus::new_default();
+        const std::string service = SyncBMCData::interface;
+        const std::string path = SyncBMCData::instance_path;
+        const std::string interface = SyncBMCData::interface;
+
+        lg2::info("datasynctool attempting a full sync.");
```
RameshIyyar, line 101: If you are using the lg2 then we should use it everywhere...you could use cerr with e.what().

src/datasynctool/dbus_interactions.cpp
```diff
@@ -28,6 +32,39 @@ PropertyMap getAllProperties(sdbusplus::bus_t& bus, const std::string& service,
     return reply.unpack<PropertyMap>();
 }
 
+int startFullSync()
+{
+    try
+    {
+        auto bus = sdbusplus::bus::new_default();
+        const std::string service = SyncBMCData::interface;
```
RameshIyyar: All these are unnecessary. I think you used it other place all ready. Again its upto you.

src/datasynctool/dbus_interactions.cpp
RameshIyyar: Agree, we should not allow to enable the sync if the redundancy is disabled since the tool could be used anytime.

src/datasynctool/dbus_interactions.cpp
aryakpadman: I feel since it's a tool, executing a single command and exit by blocking until completion is fine. 
So far no any async operation requirement is figured out, like wait for any signals and proceed.

But if you want I can make changes accordingly

src/datasynctool/dbus_interactions.cpp
aryakpadman: https://github.com/ibm-openbmc/phosphor-data-sync/pull/101#discussion_r3329543810

src/datasynctool/main.cpp
aryakpadman: CLI11 has default --help which will list the usage, hence made changes to displayy the --help output if user didn't pass any option explicitly.

src/datasynctool/main.cpp
aryakpadman: Added PeerConnected, Role and RedundancyEnabled

src/datasynctool/dbus_interactions.cpp
aryakpadman: I think you’re referring to the use of `println()`. The reason is that the data-sync application already logs detailed failure traces to the journal in many cases. For example, if a full sync fails, phosphor-data-sync logs the actual error in the journal, while datasynctool itself only reports the operation status.

If I switch these messages to `lg2`, the same information would get duplicated in the journal. Using `println()` keeps the output limited to the console and avoids redundant journal entries.

src/datasynctool/dbus_interactions.cpp
RameshIyyar: By using async you could make multiple dbus call (data sync, rbmc, and other dbus call). You could revisit it later.

src/datasynctool/dbus_interactions.cpp
```diff
@@ -0,0 +1,143 @@
...
+    try
+    {
+        auto bus = sdbusplus::bus::new_default();
+
+        auto properties = getAllProperties(bus, SyncBMCData::interface,
+                                           SyncBMCData::instance_path,
+                                           SyncBMCData::interface);
+        json statusData = buildStatusJson(properties);
+
+        auto redundancyEnabled = getProperty<bool>(
```
RameshIyyar: RedundancyEnabled and Role are hosted by the RBMC manager, why two different DBus call? you could use getAllProperties like above.

and Please use some constant and reuse it - instead of hardcoding many places

aryakpadman: Addressed by fetching all properties in a single call, and made dbus calls to async also.

RameshIyyar (approved):

PR #104, opened 2026-07-06
Implement changes for dumping all watchers info 

aryakpadman:


src/manager.cpp
```diff
@@ -922,4 +936,173 @@ sdbusplus::async::task<void> Manager::startFullSync()
     co_return;
 }
 
+void Manager::registerWatcher(const fs::path& configPath,
+                              watch::inotify::DataWatcher* watcher)
+{
+    std::lock_guard<std::mutex> lock(_watchersMutex);
```
manishtwari: The mutex are not needed here. This is a single-threaded sdbusplus event-loop all coroutines run on the same thread, we can remove _watchersMutex.

src/manager.hpp
```diff
@@ -144,6 +152,35 @@ class Manager
...
 
+    /**
+     * @brief Write all watched paths to a JSON file
+     *
+     * This method is called when SIGUSR1 signal is received. It collects all
+     * watched paths and writes them to
+     * /run/phosphor-data-sync/watching_paths.json using atomic write (temp file
+     * + rename).
+     */
+    void dumpWatchingPathsToFile() const;
```
manishtwari: can we make this private? And should `onSignalReceived` be private too?

src/rbmc_data_sync_main.cpp
```diff
@@ -49,6 +49,9 @@ int main()
         ctx, std::make_unique<data_sync::ext_data::ExternalDataIFacesImpl>(ctx),
         DATA_SYNC_CONFIG_DIR};
 
+    // Register SIGUSR1 signal handler (signal is already blocked above)
+    manager.registerSignalHandler();
```
manishtwari: We can move this to the manager constructor so `main` doesn’t need to handle it?

src/datasynctool/config_options.cpp
```diff
@@ -290,4 +296,220 @@ sdbusplus::async::task<> getPathConfig(sdbusplus::async::context& ctx,
...
+    if (kill(daemonPid, SIGUSR1) != 0)
+    {
+        std::cerr
+            << "Error: Failed to send SIGUSR1 to phosphor-data-sync (PID: "
+            << daemonPid << "): " << std::strerror(errno) << "\n";
+        co_return std::nullopt;
+    }
+
+    // Wait briefly for the daemon to write the file
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
manishtwari: 100ms looks racy. SIGUSR1 delivery does not guarantee the daemon has already rewritten watching_paths.json, so this can read a stale dump 
could we capture the current mtime, send SIGUSR1, and poll until last_write_time() changes before reading the file?

src/rbmc_data_sync_main.cpp
RameshIyyar: You could register in the Manager constructor itself.

src/manager.hpp
```diff
@@ -325,6 +395,40 @@ class Manager
...
+     * File descriptor created by signalfd() to receive SIGUSR1 signals.
+     * Set to -1 if not initialized.
+     */
+    int _signalFd{-1};
+
+    /**
+     * @brief fdio instance for signal handling
+     *
+     * Provides coroutine-based interface to monitor the signalfd for SIGUSR1.
+     * Used by onSignalReceived() coroutine to wait for signals asynchronously.
```
RameshIyyar: Please avoid mentioning function name in comment.

src/manager.hpp
```diff
@@ -325,6 +395,40 @@ class Manager
...
+     */
+    mutable std::mutex _watchersMutex;
+
+    /**
+     * @brief Signal file descriptor for SIGUSR1
+     *
+     * File descriptor created by signalfd() to receive SIGUSR1 signals.
+     * Set to -1 if not initialized.
+     */
+    int _signalFd{-1};
```
RameshIyyar: I don't think we need the `_signalFd` and `_signalFdio` member variables if `onSignalReceived` is merged into the `registerSignalHandler` function.

src/manager.hpp
```diff
@@ -325,6 +395,40 @@ class Manager
...
+     */
+    std::map<fs::path, watch::inotify::DataWatcher*> _activeWatchers;
+
+    /**
+     * @brief Mutex to protect concurrent access to _activeWatchers map
+     *
+     * This mutex ensures thread-safe registration/unregistration of watchers
+     * and safe iteration during signal-triggered dumps.
+     */
+    mutable std::mutex _watchersMutex;
```
RameshIyyar: Mutex is not required since we are running in the single thread.

src/manager.hpp
```diff
@@ -325,6 +395,40 @@ class Manager
...
+
+    /**
+     * @brief Map of config paths to their active DataWatcher instances
+     *
+     * Key: Configured path from JSON (e.g., "/var/lib/network/hypervisor/")
+     * Value: Pointer to the DataWatcher monitoring that path
+     *
+     * Protected by _watchersMutex for thread-safe access.
+     */
+    std::map<fs::path, watch::inotify::DataWatcher*> _activeWatchers;
```
RameshIyyar: Use unique pointer for DataWatcher.

src/manager.hpp
```diff
@@ -293,6 +330,39 @@ class Manager
...
+    /**
+     * @brief Register a DataWatcher with the manager.
+     *
+     * Adds a newly created DataWatcher instance to the manager's list of
+     * active watchers, allowing its lifetime to be tracked and managed.
+     *
+     * @param[in] configPath Path being monitored by the watcher.
+     * @param[in] watcher Pointer to the DataWatcher instance.
+     */
+    void registerWatcher(const fs::path& configPath,
```
RameshIyyar: You could avoid the register and unregister functions by creating the `DataWatcher` as a `std::unique_ptr` and inserting it directly into the container wherever it's created.

src/manager.hpp
```diff
@@ -144,6 +152,35 @@ class Manager
...
+    /**
+     * @brief Coroutine that waits for and handles SIGUSR1 signals
+     *
+     * This coroutine runs in a loop, waiting for SIGUSR1 signals via fdio.
+     * When a signal is received, it calls dumpWatchedPathsToFile() to write
+     * the currently watched paths to a JSON file.
+     *
+     * @return sdbusplus::async::task<> - Coroutine task
+     */
+    sdbusplus::async::task<> onSignalReceived();
```
RameshIyyar: It could be into private scope and also you remove this function as mentioned in other comment.

src/manager.hpp
```diff
@@ -144,6 +152,35 @@ class Manager
...
+     */
+    void dumpWatchingPathsToFile() const;
+
+    /**
+     * @brief Register SIGUSR1 signal handler using signalfd
+     *
+     * Sets up signalfd to receive SIGUSR1 signals and creates an fdio instance
+     * to monitor it. Spawns the onSignalReceived() coroutine to handle signals.
+     */
+    void registerSignalHandler();
```
RameshIyyar: Its not required into public scope if you are calling this function into constructor as mentioned in another comment.

src/manager.hpp
RameshIyyar: It need to be in private scope.

src/manager.hpp
```diff
@@ -144,6 +152,35 @@ class Manager
      */
     void setSyncEventsHealth(const SyncEventsHealth& syncEventsHealth);
 
+    /**
+     * @brief Write all watched paths to a JSON file
+     *
+     * This method is called when SIGUSR1 signal is received. It collects all
+     * watched paths and writes them to
+     * /run/phosphor-data-sync/watching_paths.json using atomic write (temp file
```
RameshIyyar: Please avoid capturing path info in the comment.

src/manager.cpp
```diff
@@ -922,4 +936,173 @@ sdbusplus::async::task<void> Manager::startFullSync()
...
+        lg2::error("Failed to read from signalfd: {ERROR}", "ERROR",
+                   std::strerror(errno));
+    }
+
+    co_return;
+}
+
+void Manager::dumpWatchingPathsToFile() const
+{
+    constexpr auto outputFile = "/run/phosphor-data-sync/watching_paths.json";
```
RameshIyyar: Please use `/tmp` and lets not use directory, just use filename as `data-sync_watching_paths.json`.

src/manager.cpp
```diff
@@ -922,4 +936,173 @@ sdbusplus::async::task<void> Manager::startFullSync()
...
+
+    try
+    {
+        fs::path dirPath = fs::path(outputFile).parent_path();
+
+        if (!fs::exists(dirPath))
```
RameshIyyar: If we used tmp directory alone then we don't need this block.

src/manager.cpp
```diff
@@ -922,4 +936,173 @@ sdbusplus::async::task<void> Manager::startFullSync()
...
+        {
+            fs::create_directories(dirPath);
+        }
+
```
RameshIyyar: Please add logic to remove `data-sync_watching_paths.json` file if exist and then dump data into new file.

src/manager.cpp
```diff
@@ -922,4 +936,173 @@ sdbusplus::async::task<void> Manager::startFullSync()
...
+        nlohmann::json output = collectAllWatchingPaths();
+
+        // Write atomically to the file by using temp file + rename
+        std::string tempFile = std::string(outputFile) + ".tmp";
```
RameshIyyar: Its not required.

src/manager.cpp
```diff
@@ -922,4 +936,173 @@ sdbusplus::async::task<void> Manager::startFullSync()
...
+    try
+    {
+        // Create fdio instance to monitor the signalfd
+        _signalFdio = std::make_unique<sdbusplus::async::fdio>(_ctx, _signalFd);
+
+        _ctx.spawn([](sdbusplus::async::context& ctx,
+                      Manager* mgr) -> sdbusplus::async::task<> {
+            while (!ctx.stop_requested())
+            {
+                co_await mgr->onSignalReceived();
```
RameshIyyar: As mentioned in other place, move `onSignalReceived` and remove this co_await.

src/manager.cpp
```diff
@@ -922,4 +936,173 @@ sdbusplus::async::task<void> Manager::startFullSync()
...
+
+        lg2::info(
+            "Successfully registered SIGUSR1 handler using fdio (fd={FD})",
+            "FD", _signalFd);
+    }
+    catch (const std::exception& e)
+    {
+        lg2::error("Failed to register SIGUSR1 handler: {ERROR}", "ERROR",
+                   e.what());
+        close(_signalFd);
```
RameshIyyar: You should only if its not `-1`

src/manager.cpp
```diff
@@ -922,4 +936,173 @@ sdbusplus::async::task<void> Manager::startFullSync()
...
+        }
+    }
+}
+
+nlohmann::json Manager::collectAllWatchingPaths() const
+{
+    nlohmann::json result;
+    nlohmann::json watchingPaths;
+
+    std::lock_guard<std::mutex> lock(_watchersMutex);
```
RameshIyyar: Mutes is not required as mentioned in other comment.

src/manager.cpp
```diff
@@ -922,4 +936,173 @@ sdbusplus::async::task<void> Manager::startFullSync()
...
+
+    lg2::debug("Collecting the {COUNT} active watchers", "COUNT",
+               _activeWatchers.size());
+
+    for (const auto& [configPath, dataWatcher] : _activeWatchers)
```
RameshIyyar, line 957: Please std::ranges algorithm.

src/manager.cpp
aryakpadman: Agree, Removed the registerWatcher and unregisterWatcher APIs as well and made the logic inline

src/manager.cpp
aryakpadman: Acknowledged

src/manager.cpp
aryakpadman: Done

src/manager.hpp
aryakpadman: Ack.

src/manager.hpp
aryakpadman: Done

src/manager.hpp
aryakpadman: Acknowledged

src/manager.hpp
aryakpadman: Addressed.

src/manager.hpp
aryakpadman: Yeah agree. Made into local scope

src/manager.cpp
aryakpadman: got it. For ease of handling FD changed the type to utils::FD, and same is updated with move constructor

src/manager.cpp
aryakpadman: used std::ranges::transform inside and outer loop kept as it is for better readability

src/manager.cpp
```diff
@@ -922,4 +944,134 @@ sdbusplus::async::task<void> Manager::startFullSync()
...
+
+        utility::FD sigusr1Fd{signalfd(-1, &ss, SFD_NONBLOCK | SFD_CLOEXEC)};
+        if (sigusr1Fd() < 0)
+        {
+            lg2::error("Failed to create signalfd: {ERROR}", "ERROR",
+                       std::strerror(errno));
+            return;
+        }
+
+        lg2::info(
```
spinler: lg2::debug?

src/datasynctool/config_options.cpp
aryakpadman: I went with the sleep approach since is a tool only and doesnt impact the production code. 
Thinking another approach as well that leverages the inotify via existing DataWatcher framework of the data-sync code, which I believe inotify would be more appropriate solution here.However still need to find a way to disable or redirect the logger traces of the DataWatcher framework from the tool output in console

For the time being, I'll add a TODO to revisit this improvement and increase the sleep duration as a temporary workaround.

src/manager.cpp
aryakpadman: Done

manishtwari (approved):

src/datasynctool/config_options.cpp
```diff
@@ -290,4 +295,227 @@ sdbusplus::async::task<> getPathConfig(sdbusplus::async::context& ctx,
...
+            result["config_path"] = foundConfigPath;
+        }
+        std::println("{}", result.dump(4));
+    }
+    else
+    {
+        if (found)
+        {
+            std::println("Path[{}] is watching currently.", targetPath);
+            std::println("  Config Path : {}", foundConfigPath);
```
RameshIyyar, line 394: nit: Remove space before `Config`. If you added intentionally then ignore.

RameshIyyar (approved):

spinler (approved):

src/datasynctool/config_options.cpp
aryakpadman: Yes, it was intentional

PR #106, opened 2026-07-13
Datasynctool: Adding options to get sync failure data 

riya461:
- Added a new option `--error-logs` or `-E` to get the summary of all pels under `bmc datasync` component.

### DIFF
--- a/src/datasynctool/error_summary.cpp
+++ b/src/datasynctool/error_summary.cpp
@@ -1,0 +1,226 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#include "error_summary.hpp"
+
+#include <sys/wait.h>
+
+#include <nlohmann/json.hpp>
+
+#include <array>
+#include <cerrno>
+#include <cstdio>
+#include <format>
+#include <map>
+#include <optional>
+#include <print>
+#include <ranges>
+#include <span>
+#include <string>
+#include <string_view>
+#include <system_error>
+#include <vector>
+
+namespace datasynctool::error_summary
+{
+
+using json = nlohmann::ordered_json;
+
+// ── Constants
+// ─────────────────────────────────────────────────────────────────
+
+constexpr std::string_view datasyncSrcPrefix = "BD8D70";
+
+constexpr std::string_view separator =
+    "─────────────────────────────────────────────────";
+
+// Read buffer for popen output
+constexpr std::size_t readBufSize = 4096;
+
+// Registry message map
+static const std::map<std::string_view, std::string_view> srcRegistryMap = {
+    {"BD8D7001", "SyncFailure"},
+    {"BD8D7002", "SyncEventsFailure"},
+    {"BD8D7003", "ParserFailure"},
+    {"BD8D7004", "NotifyFailure"},
+};
+
+// Field descriptors
+static constexpr PelField fieldRegistryMsg = {"Primary SRC", "Reference Code",
+                                              "Registry Message"};
+
+static constexpr PelField fieldFailureTime = {"Private Header", "Committed at",
+                                              "Failure Time"};
+
+static constexpr PelField fieldPelId = {"Private Header", "Platform Log Id",
+                                        "PEL ID"};
+
+// SyncFailure - only fields
+static constexpr PelField fieldPath = {"User Data 1", "DS_Sync_Path", "Path"};
+static constexpr PelField fieldErrMsg = {"User Data 1", "DS_Sync_ErrMsg",
+                                         "ErrMsg"};
+static constexpr PelField fieldErrCode = {"User Data 1", "DS_Sync_ErrCode",
+                                          "ErrCode"};
+
+static std::optional<std::string> runCommand(std::string_view cmd)
+{
+    // NOLINTNEXTLINE
+    FILE* pipe = popen(std::string(cmd).c_str(), "r");
+    if (pipe == nullptr)
+    {
+        std::println(stderr, "popen failed for command: {}", cmd);
+        return std::nullopt;
+    }
+
+    std::string output;
+    output.reserve(readBufSize);
+    std::array<char, readBufSize> buf{};
+    std::size_t n = 0;
+    while ((n = std::fread(buf.data(), 1, buf.size(), pipe)) > 0)
+    {
+        const auto chunk = std::span(buf).first(n);
+        output.append(chunk.begin(), chunk.end());
+    }
+
+    const int status = pclose(pipe);
+    if (status == -1)
+    {
+        std::println(stderr, "pclose failed: {}",
+                     std::system_category().message(errno));
+    }
+    else if (WIFEXITED(status) && WEXITSTATUS(status) != 0)
+    {
+        std::println(stderr, "peltool.py exited with code {}",
+                     WEXITSTATUS(status));
+    }
+
+    return output;
+}
+
+// Extract a string value from a PEL JSON object using a PelField descriptor.
+// Returns an empty string if the parent section or child key is absent.
+static std::string_view extractField(const json& pelData, const PelField& field)
+{
+    const auto parentIt = pelData.find(field.parent);
+    if (parentIt == pelData.end() || !parentIt->is_object())
+    {
+        return {};
+    }
+    const auto childIt = parentIt->find(field.child);
+    return (childIt != parentIt->end() && childIt->is_string())
+               ? std::string_view(childIt->get_ref<const std::string&>())
+               : std::string_view{};
+}
+
+// Build a SummaryEntry from a single PEL JSON object.
+static SummaryEntry makeSummaryEntry(const json& pelData)
+{
+    const std::string_view rawSrc = extractField(pelData, fieldRegistryMsg);
+    const auto it = srcRegistryMap.find(rawSrc);
+    // Resolve the raw SRC reference code to a registry message string.
+    // Falls back to the raw code if not found in the map.
+    const std::string_view regMsg = it != srcRegistryMap.end() ? it->second
+                                                               : rawSrc;
+
+    SummaryEntry entry{
+        .registryMsg = std::string(regMsg),
+        .failureTime = std::string(extractField(pelData, fieldFailureTime)),
+        .pelId = std::string(extractField(pelData, fieldPelId)),
+        .path = {},
+        .errMsg = {},
+        .errCode = {},
+    };
+
+    if (regMsg == "SyncFailure")
+    {
+        entry.path = std::string(extractField(pelData, fieldPath));
+        entry.errMsg = std::string(extractField(pelData, fieldErrMsg));
+        entry.errCode = std::string(extractField(pelData, fieldErrCode));
+    }
+
+    return entry;
+}
+
+sdbusplus::async::task<> displayErrorLogSummary(bool jsonOutput)
+{
+    const auto output = runCommand(
+        std::format("peltool.py --src {} -r 1 -a", datasyncSrcPrefix));
+
+    if (!output.has_value() || output->empty())
+    {
+        std::println("peltool.py returned no output for SRC prefix {}",
+                     datasyncSrcPrefix);
+        co_return;
+    }
+
+    json pelMap;
+    try
+    {
+        pelMap = json::parse(*output);
+    }
+    catch (const json::exception& e)
+    {
+        std::println(stderr, "Failed to parse peltool output: {}", e.what());
+        co_return;
+    }
+
+    if (!pelMap.is_object())
+    {
+        std::println(stderr, "Unexpected peltool output format");
+        co_return;
+    }
+
+    // Collect one SummaryEntry per error log.
+    auto entries = pelMap.items() | std::views::filter([](const auto& kv) {
+        return kv.value().is_object();
+    }) | std::views::transform([](const auto& kv) {
+        return makeSummaryEntry(kv.value());
+    }) | std::views::filter([](const SummaryEntry& e) {
+        return !e.registryMsg.empty() || !e.failureTime.empty() ||
+               !e.pelId.empty();
+    }) | std::ranges::to<std::vector>();
+
+    if (entries.empty())
+    {
+        std::println("No DataSync error logs found");
+        co_return;
+    }
+
+    if (jsonOutput)
+    {
+        auto out = entries | std::views::transform([](const SummaryEntry& e) {
+            json obj;
+            obj[fieldRegistryMsg.displayName] = e.registryMsg;
+            obj[fieldFailureTime.displayName] = e.failureTime;
+            obj[fieldPelId.displayName] = e.pelId;
+            if (!e.path.empty() || !e.errMsg.empty() || !e.errCode.empty())
+            {
+                obj[fieldPath.displayName] = e.path;
+                obj[fieldErrMsg.displayName] = e.errMsg;
+                obj[fieldErrCode.displayName] = e.errCode;
+            }
+            return obj;
+        }) | std::ranges::to<json>();
+        std::println("{}", out.dump(2));
+        co_return;
+    }
+
+    // Text output
+    for (const auto& e : entries)
+    {
+        std::println("{}", separator);
+        std::println("  {:<18}: {}", fieldRegistryMsg.displayName,
+                     e.registryMsg);
+        std::println("  {:<18}: {}", fieldFailureTime.displayName,
+                     e.failureTime);
+        std::println("  {:<18}: {}", fieldPelId.displayName, e.pelId);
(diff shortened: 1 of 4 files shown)


src/datasynctool/utils.cpp
```diff
@@ -2,11 +2,115 @@
...
 
+#include <array>
+#include <format>
 #include <print>
+#include <stdexcept>
 
 namespace datasynctool::utils
 {
 
+sdbusplus::async::task<std::pair<int, std::string>>
```
aryakpadman: This code seems duplicating the [AsyncCommandExecutor](https://github.com/ibm-openbmc/phosphor-data-sync/blob/main/src/async_command_exec.hpp) logic .

We can import and use the existing AsyncCommandExecutor from src/.

Or else we can keep the common codes like async_command_exec.hpp and utils.cpp in a common space and can use like a shared library.

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,247 @@
...
+    std::pair{"DS_Sync_"sv, "SyncFailure"sv},
+    std::pair{"DS_Notify_"sv, "NotifyFailure"sv},
+    std::pair{"DS_Parser_"sv, "ParserFailure"sv},
+    std::pair{"DS_Events_"sv, "SyncEventsFailure"sv},
+};
+
+/**
+ * @brief Return a string field or an empty view.
+ */
+std::string_view strViewOrEmpty(const json& j, std::string_view key)
```
aryakpadman: Can we make these APIs static?

src/datasynctool/utils.cpp
```diff
@@ -2,11 +2,115 @@
...
+    // NOLINTNEXTLINE
+    CommandExecutor::execCmd(const std::string& cmd)
```
manishtwari: Can we reuse the existing `data_sync::async::AsyncCommandExecutor` and `data_sync::utility::FD` instead of adding another CommandExecutor/FD under `datasynctool::utils`?

src/datasynctool/main.cpp
```diff
@@ -75,6 +85,16 @@ int main(int argc, char* argv[])
 
     sdbusplus::async::context ctx;
 
+    if (showErrorLog)
+    {
+        ctx.spawn(datasynctool::error_summary::displayErrorLogSummary(
+                      ctx, jsonOutput) |
+                  sdbusplus::async::execution::then(
+                      [&ctx]() { ctx.request_stop(); }));
+        ctx.run();
```
manishtwari: Do we need this separate `ctx.run()` and `return` path for -E, or can we spawn `displayErrorLogSummary()` like the other actions and let the common stop block handle it?

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,247 @@
...
+    }
+    catch (const std::exception& e)
+    {
+        std::println(stderr, "{}", e.what());
+        co_return;
+    }
+
+    auto& [exitCode, output] = result;
+    if (exitCode != 0)
+    {
```
manishtwari: Should we print an error here instead of silently returning?

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,247 @@
...
+        co_return;
+    }
+
+    if (output.empty())
+    {
+        co_return;
```
aryakpadman, line 191: In this case the tool will return empty right? Feel we could have a meaningful message as well.

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+    if (jsonOutput)
+    {
+        std::println("{}", failures.dump(2));
+        co_return;
+    }
+
+    std::println("\nData-sync failure error logs ({} found):\n",
+                 failures.size());
+    for (const auto& entry : failures)
```
aryakpadman: nit : we could use std::ranges::for_each() here

src/datasynctool/utils.cpp
riya461: Updating to use it as a shared library. I haven't changed to a common space but included those directories.

src/datasynctool/error_summary.cpp
riya461: In this case 
`'strViewOrEmpty'` is a static definition in anonymous namespace, static is redundant here 
`readability-static-definition-in-anonymous-namespace` error will hit

src/datasynctool/main.cpp
riya461: In case of other actions, they are all D-Bus calls, none of which check stop_requested(). Due to the `AsyncCommandExecutor`, -E is stop-sensitive.

src/datasynctool/utils.cpp
```diff
@@ -3,6 +3,7 @@
 #include "utils.hpp"
 
 #include <print>
+#include <string>
```
manishtwari: is it required?

src/datasynctool/main.cpp
manishtwari: Got it, -E needs separate ctx.run() because Async command checks stop_requested()
my question is about behavior when user gives multiple options for example, datasynctool -E -s is accepted today, but only -E runs and -s is skipped because of the early return, should we add excludes() for other options so CLI rejects this clearly?

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+} // namespace
+
+sdbusplus::async::task<> displayErrorLogSummary(sdbusplus::async::context& ctx,
+                                                bool jsonOutput,
+                                                std::size_t maxLogs,
+                                                bool verbose)
+{
+    const std::string cmd = std::format(
+        "peltool.py --src {} -r{} -a 2>/dev/null", datasyncSrcPrefix,
```
manishtwari: do we still need `2>/dev/null` here? AsyncCommandExecutor already captures stderr. If we remove this, we can see the actual peltool.py error right?

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,248 @@
+// SPDX-License-Identifier: Apache-2.0
```
RameshIyyar, line 1: A couple of high level Qs, 

- I thought we had discussed adding an option to display only the last sync failure details. However, in this commit, we're displaying all sync failures in reverse order. Is that intentional?
- Why do we need to customize the peltool output? We could simply retrieve the required fields from the PEL JSON output and print them in our own JSON format. There's no need to suppress or reformat most of the existing peltool details. Reusing the peltool JSON output (with only the required fields filtered) would simplify the implementation significantly and allow us to avoid the additional error_summary handling.
- Why do we need to run the command asynchronously? This is a CLI tool, and this option doesn't appear to perform any other work in parallel or require non-blocking execution. Wouldn't a simple synchronous system() call (or equivalent) make the implementation much simpler?

Sync failure could be simplified

```
{
   "ErrCode" : 12,
   "ErrMsg"   : "rsync: [sender] safe_read failed to read 1 bytes: Connection reset by peer (104)
rsync error: error in rsync protocol data stream (code 12) at io.c(283) [sender=3.4.1]",
   "Failure_Time: "07/09/2026 15:09:41",
   "Path" : "/var/lib/phosphor-logging/errors/1308",
   "PEL_ID" : "0x50000716",
} 
```

If I passed --sync_failure 10 then last 10 failure.
If I passed --sync_failure --trace then show with traces (data sync and stunnel).
If I passed --sync_failure 10 --trace then show last 10 sync failure with traces (data sync and stunnel).

src/datasynctool/main.cpp
aryakpadman: Is the code really hitting `ctx.run() and stop_requested()` at the end of the main.cpp ?

src/datasynctool/utils.cpp
aryakpadman: ok

src/datasynctool/main.cpp
riya461: Since the implementation is simpler in synchronous, this code and logic would be changed.

src/datasynctool/error_summary.cpp
riya461: Updated based on the above suggestion.

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,227 @@
...
```
aryakpadman, line 1: commit message : ` summary of recent pel` , not  `summary of all pels`, Right?

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+#include "error_summary.hpp"
+
+#include <sys/wait.h>
+
+#include <nlohmann/json.hpp>
+
+#include <algorithm>
```
manishtwari: nit: not used algorithm can be removed.

ARYAPADMAN (approved):

aryakpadman (approved):

manishtwari (approved):

spinler (approved):

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,226 @@
...
+constexpr std::string_view datasyncSrcPrefix = "BD8D70";
+
+constexpr std::string_view separator =
+    "─────────────────────────────────────────────────";
+
+// Read buffer for popen output
+constexpr std::size_t readBufSize = 4096;
+
+// Registry message map
+static const std::map<std::string_view, std::string_view> srcRegistryMap = {
```
RameshIyyar, line 40: Why do we need this map? Can't we simply read the Message field defined in the registry configuration file?

This would require updating this list whenever a new data registry is added, which creates a maintenance burden.

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,226 @@
...
+               ? std::string_view(childIt->get_ref<const std::string&>())
+               : std::string_view{};
+}
+
+// Build a SummaryEntry from a single PEL JSON object.
+static SummaryEntry makeSummaryEntry(const json& pelData)
+{
+    const std::string_view rawSrc = extractField(pelData, fieldRegistryMsg);
+    const auto it = srcRegistryMap.find(rawSrc);
+    // Resolve the raw SRC reference code to a registry message string.
```
RameshIyyar, line 119: The same comment (that I posted in this map initialization) is applicable to here.

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,226 @@
...
+    SummaryEntry entry{
+        .registryMsg = std::string(regMsg),
+        .failureTime = std::string(extractField(pelData, fieldFailureTime)),
+        .pelId = std::string(extractField(pelData, fieldPelId)),
+        .path = {},
+        .errMsg = {},
+        .errCode = {},
+    };
+
+    if (regMsg == "SyncFailure")
```
RameshIyyar, line 133: Couldn't we simply print the field if it exists in the configured PEL? Why do we need registry-specific checks?

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,226 @@
...
+static SummaryEntry makeSummaryEntry(const json& pelData)
+{
+    const std::string_view rawSrc = extractField(pelData, fieldRegistryMsg);
+    const auto it = srcRegistryMap.find(rawSrc);
+    // Resolve the raw SRC reference code to a registry message string.
+    // Falls back to the raw code if not found in the map.
+    const std::string_view regMsg = it != srcRegistryMap.end() ? it->second
+                                                               : rawSrc;
+
+    SummaryEntry entry{
```
RameshIyyar, line 124: Instead, we could define a vector of PelField objects and iterate over it to extract and print the corresponding fields from the PEL. This way, we don't need to add separate extraction logic for every new field also we could avoid SummaryEntry.

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    if (!pelMap.is_object())
+    {
+        std::println(stderr, "Unexpected peltool output format");
+        co_return;
+    }
+
+    // Collect one SummaryEntry per error log.
+    auto entries = pelMap.items() | std::views::filter([](const auto& kv) {
```
RameshIyyar, line 173: I think its always return object so do we really need this?

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,226 @@
...
+        return kv.value().is_object();
+    }) | std::views::transform([](const auto& kv) {
+        return makeSummaryEntry(kv.value());
+    }) | std::views::filter([](const SummaryEntry& e) {
```
RameshIyyar, line 177: All these fields are mandatory field in the PEL so I dont think we will have case where we wont this field.

src/datasynctool/main.cpp
```diff
@@ -57,6 +57,13 @@ int main(int argc, char* argv[])
             ->expected(0, 1)
             ->default_val(1);
 
+    bool includeTrace{false};
+    auto* traceOpt = errorGroup->add_flag(
+        "-T,--trace", includeTrace,
```
RameshIyyar, line 62: can we use small `t` for short options?

src/datasynctool/error_summary.cpp
```diff
@@ -0,0 +1,226 @@
...
+// Registry message map
+static const std::map<std::string_view, std::string_view> srcRegistryMap = {
+    {"BD8D7001", "SyncFailure"},
+    {"BD8D7002", "SyncEventsFailure"},
+    {"BD8D7003", "ParserFailure"},
+    {"BD8D7004", "NotifyFailure"},
+};
+
+// Field descriptors
+static constexpr PelField fieldRegistryMsg = {"Primary SRC", "Reference Code",
```
RameshIyyar, line 48: nlohmann JSON supports [pointer](https://json.nlohmann.me/features/json_pointer/) so that you could form a key (includes nested keys) in string format and extract value

src/datasynctool/error_summary.cpp
```diff
@@ -111,8 +111,39 @@ static std::string_view extractField(const json& pelData, const PelField& field)
                : std::string_view{};
 }
 
+// Extract trace lines from "User Data 2" and "User Data 3" "Data" arrays.
+static std::vector<std::string> extractTraceLines(const json& pelData)
+{
+    std::vector<std::string> lines;
+
+    for (const std::string_view section : {"User Data 2", "User Data 3"})
```
RameshIyyar, line 119: You could use the nlohmann JSON [pointer](https://json.nlohmann.me/features/json_pointer/) approach to extract the traces. I believe UD2 is for Data Sync and UD3 is for stunnel, so it would be better to print them separately rather than combining them. Since their timestamps can differ, keeping them separate makes it easier to correlate the logs based on the timestamps.

src/datasynctool/error_summary.cpp
RameshIyyar: We could address this in the next commit series, its not affecting any functionality

src/datasynctool/error_summary.cpp
RameshIyyar: We could address this in the next commit series, its not affecting any functionality.

src/datasynctool/error_summary.cpp
RameshIyyar: We could address this in the next commit series, its not affecting any functionality.

src/datasynctool/error_summary.cpp
RameshIyyar: We could address this in the next commit series, its not affecting any functionality.

src/datasynctool/error_summary.cpp
RameshIyyar: We could address this in the next commit series, its not affecting any functionality.

src/datasynctool/error_summary.cpp
RameshIyyar: We could address this in the next commit series, its not affecting any functionality.

src/datasynctool/error_summary.cpp
RameshIyyar: We could address this in the next commit series, its not affecting any functionality.

src/datasynctool/error_summary.cpp
RameshIyyar: We could address this in the next commit series, its not affecting any functionality.

src/datasynctool/main.cpp
RameshIyyar: We could address this in the next commit series, its not affecting any functionality.

RameshIyyar (approved):
Most of my comments are about code simplification rather than functionality. We can address them in the next commit series, where we already plan to simplify how error details are printed.

PR #108, opened 2026-07-17
ExternalData: Read BMC position from runtime file

manishtwari:
The local BMC position is now provided through a platform runtime file so it can be consumed from the common source.

### DIFF
--- a/src/external_data_ifaces_impl.cpp
+++ b/src/external_data_ifaces_impl.cpp
@@ -5,47 +5,20 @@
 #include "error_log.hpp"
 
 #include <phosphor-logging/lg2.hpp>
-#include <xyz/openbmc_project/Inventory/Decorator/Position/client.hpp>
 #include <xyz/openbmc_project/Logging/Create/client.hpp>
-#include <xyz/openbmc_project/ObjectMapper/client.hpp>
 #include <xyz/openbmc_project/State/BMC/Redundancy/client.hpp>
+
+#include <fstream>
+#include <limits>
 
 namespace data_sync::ext_data
 {
 
+constexpr auto bmcPositionFile = "/run/openbmc/bmc_position";
+
 ExternalDataIFacesImpl::ExternalDataIFacesImpl(sdbusplus::async::context& ctx) :
     _ctx(ctx)
 {}
-
-sdbusplus::async::task<std::string>
-    // NOLINTNEXTLINE
-    ExternalDataIFacesImpl::getDBusService(const std::string& objPath,
-                                           const std::string& interface)
-{
-    try
-    {
-        using ObjectMapperMgr =
-            sdbusplus::client::xyz::openbmc_project::ObjectMapper<>;
-
-        auto objectMapperMgr = ObjectMapperMgr(_ctx)
-                                   .service(ObjectMapperMgr::default_service)
-                                   .path(ObjectMapperMgr::instance_path);
-
-        std::vector<std::string> interfaces{interface};
-
-        auto services = co_await objectMapperMgr.get_object(objPath,
-                                                            interfaces);
-
-        co_return services.begin()->first;
-    }
-    catch (const std::exception& e)
-    {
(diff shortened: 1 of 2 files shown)


riya461 (approved):

aryakpadman (approved):

src/external_data_ifaces_impl.hpp
```diff
@@ -60,26 +60,13 @@ class ExternalDataIFacesImpl : public ExternalDataIFaces
...
-                       const std::string& interface);
-
     /**
      * @brief  Used to retrieve the BMC role from DBus.
      */
     sdbusplus::async::task<> fetchBMCRedundancyMgrProps() override;
 
     /**
-     * @brief Used to retrieve the BMC Position from Dbus.
+     * @brief Used to retrieve the BMC Position from from the runtime file.
```
RameshIyyar: typo: `from from`

src/external_data_ifaces_impl.cpp
```diff
@@ -1,52 +1,24 @@
 // SPDX-License-Identifier: Apache-2.0
 
+#include "config.h"
```
RameshIyyar: Commit message:

The purpose of this change is not to avoid BMC VPD collection.

Ideally, the BMC position should be exposed through `Item.BMC` inventory object. However, on some platforms (for example, IBM systems), the BMC position is required before BMC VPD can be collected. Since BMC VPD collection occurs only after role determination on some platforms, the BMC position must be available earlier in the BMC boot process.

Currently, the BMC position is obtained from the System inventory, but this is an IBM-specific implementation and should not be treated as a generic solution. The OpenBMC community recommended maintaining the running BMC position in `/run/openbmc`, with the mechanism for determining the BMC position remaining platform-specific.

This change updates the code to read the running BMC position from `/run/openbmc/bmc_position` instead of relying on the System inventory, making the implementation platform-independent.

RameshIyyar (approved):

src/external_data_ifaces_impl.cpp
```diff
@@ -77,22 +49,23 @@ sdbusplus::async::task<> ExternalDataIFacesImpl::fetchBMCPosition()
...
+
+        BMCPosition pos{};
+        if (!(posFile >> pos))
+        {
+            throw std::runtime_error(
+                std::string("Invalid BMC position value in ") +
+                BMC_POSITION_FILE);
+        }
+
+        bmcPosition(pos);
```
RameshIyyar: If the BMC position is invalid, the application should terminate.

I would prefer not to create an error log when the BMC position state is invalid, since another application is responsible for reading the BMC position from the cable and updating it in both D-Bus and the filesystem. We can rely on that application to report any error log related to the BMC position.

src/external_data_ifaces_impl.cpp
```diff
@@ -75,31 +47,24 @@ sdbusplus::async::task<> ExternalDataIFacesImpl::fetchBMCRedundancyMgrProps()
...
-                        .service(service)
-                        .path(systemInvObjPath)
-                        .position());
+        throw std::runtime_error(std::string("Cannot open ") +
+                                 BMC_POSITION_FILE);
     }
-    catch (const std::exception& e)
+
+    BMCPosition pos{};
+    // VPD writes max<size_t>() when the BMC position cannot be determined.
```
RameshIyyar: Please avoid mention about IBM component names in the common repo.

meson.build
```diff
@@ -119,6 +119,12 @@ conf_data.set_quoted(
     description: 'BMC1 rsyncd port',
 )
 
+conf_data.set_quoted(
+    'BMC_POSITION_FILE',
```
RameshIyyar: Do we need to set from meson?

manishtwari: No, it does not need to come from meson since the path is fixed and not platform configurable. I’ll define it as a constexpr in the source file instead.

RameshIyyar (approved):

spinler (approved):

PR #110, opened 2026-07-23
manager: avoid duplicate bidirectional watchers on sync enable

manishtwari:
During failover, the RBMC state manager disables background sync and then enables it again before starting a full sync. Data sync currently does not cancel active inotify watchers when sync is disabled.

For bidirectional immediate sync configs, the existing watcher is still valid before and after the role change. When sync is enabled again, data sync can try to start another watcher for the same path that can result in two awaits on the same fdio object and crash with:
```
 what():fdio_completion started with another await already pending!
```
Skip starting a bidirectional immediate watcher when a watcher for the same configured path is already active. This avoids the duplicate watcher during failover while still allowing fresh startup to create the watcher normally.

### DIFF
--- a/src/manager.cpp
+++ b/src/manager.cpp
@@ -264,9 +264,21 @@ sdbusplus::async::task<> Manager::startSyncEvents()
         return this->isSyncEligible(dataSyncCfg);
     }),
         [this](const auto& dataSyncCfg) {
+        using enum config::SyncDirection;
         using enum config::SyncType;
         if (dataSyncCfg._syncType == Immediate)
         {
+            // Bidirectional watchers can already be running when sync is
+            // enabled again during failover. Skip creating the same watcher
+            // again.
+            if ((dataSyncCfg._syncDirection == Bidirectional) &&
+                _activeWatchers.contains(dataSyncCfg._path))
+            {
+                lg2::debug(
+                    "Bidirectional watcher already exists for {PATH}, skipping duplicate watcher",
+                    "PATH", dataSyncCfg._path);
+                return;
+            }
             try
             {
                 this->_ctx.spawn(this->monitorDataToSync(dataSyncCfg));


src/manager.cpp
```diff
@@ -264,9 +264,21 @@ sdbusplus::async::task<> Manager::startSyncEvents()
...
         using enum config::SyncType;
         if (dataSyncCfg._syncType == Immediate)
         {
+            // Bidirectional watchers can already be running when sync is
+            // enabled again during failover. Skip creating the same watcher
+            // again.
+            if ((dataSyncCfg._syncDirection == Bidirectional) &&
+                _activeWatchers.contains(dataSyncCfg._path))
+            {
+                lg2::debug(
```
RameshIyyar, line 277: Do we need debug trace?

src/manager.cpp
```diff
@@ -264,9 +264,21 @@ sdbusplus::async::task<> Manager::startSyncEvents()
...
         [this](const auto& dataSyncCfg) {
+        using enum config::SyncDirection;
         using enum config::SyncType;
         if (dataSyncCfg._syncType == Immediate)
         {
+            // Bidirectional watchers can already be running when sync is
+            // enabled again during failover. Skip creating the same watcher
+            // again.
+            if ((dataSyncCfg._syncDirection == Bidirectional) &&
+                _activeWatchers.contains(dataSyncCfg._path))
```
RameshIyyar, line 275: LGTM but its not ideal solution.

What happens if reset this container objects? does it remove all active watchers? I guess no, could you please check it? if its working then lets go with that fix so that we will remove and create watchers whenever sync is disabled (remove watcher - ongoing sync would be interrupted?) and enabled(create watchers)

RameshIyyar (approved):

src/manager.cpp
RameshIyyar: If its not fixing then we will go with this fix for now and find a proper way to remove and create watchers during sync disable/enable.

manishtwari: activeWatchers owns unique_ptr<DataWatcher>, clearing it should destroy the DataWatcher objects, remove the inotify watches, and close the fd.

but I need to check if it is safe in the current flow monitorDataToSync keeps a raw DataWatcher* and may be waiting inside onDataChange(), if we clear the container while that coroutine is still using the watcher, it may become unsaf i will check this path
Also, removing the watcher should not interrupt an already-running rsync it should only stop future inotify events.

src/manager.cpp
manishtwari: I agree, it’s not really required.

RameshIyyar (approved):

aryakpadman (approved):

PR #111, opened 2026-07-30
Add deferred sync support

manishtwari:
This PR adds Deferred sync support.

In Deferred sync, data-sync waits for a short configured delay after a file change, then runs one sync for the configured path. This helps reduce the number of rsync commands during bulk file changes.

### DIFF
--- a/config/schema/example.json
+++ b/config/schema/example.json
@@ -55,6 +55,13 @@
                 "Mode": "DBus",
                 "NotifyServices": ["Service1", "Service2"]
             }
+        },
+        {
+            "Path": "/directory3/path/to/sync/",
+            "Description": "Add details about the data and purpose of the synchronization",
+            "SyncDirection": "Bidirectional",
+            "SyncType": "Deferred",
+            "DeferredSyncInterval": "PT1S"
         }
     ]
 }
--- a/src/data_sync_config.cpp
+++ b/src/data_sync_config.cpp
@@ -75,6 +75,19 @@ DataSyncConfig::DataSyncConfig(const nlohmann::json& config,
         _periodicityInSec = std::nullopt;
     }
 
+    if (_syncType == SyncType::Deferred)
+    {
+        constexpr auto defDeferredSyncInterval = 1;
+        _deferredSyncIntervalInSec =
+            convertISODurationToSec(
+                config["DeferredSyncInterval"].get<std::string>())
+                .value_or(std::chrono::seconds(defDeferredSyncInterval));
+    }
+    else
+    {
+        _deferredSyncIntervalInSec = std::nullopt;
+    }
+
     if (config.contains("NotifySibling"))
     {
         _notifySibling = NotifySiblingConfig(config["NotifySibling"]);
@@ -123,6 +136,8 @@ bool DataSyncConfig::operator==(const DataSyncConfig& dataSyncCfg) const
            _destPath == dataSyncCfg._destPath &&
            _syncType == dataSyncCfg._syncType &&
            _periodicityInSec == dataSyncCfg._periodicityInSec &&
+           _deferredSyncIntervalInSec ==
+               dataSyncCfg._deferredSyncIntervalInSec &&
            _retry == dataSyncCfg._retry &&
            _excludeList == dataSyncCfg._excludeList &&
            _includeList == dataSyncCfg._includeList;
@@ -172,6 +187,10 @@ std::optional<SyncType>
     if (syncType == "Immediate")
     {
         return SyncType::Immediate;
+    }
+    else if (syncType == "Deferred")
+    {
+        return SyncType::Deferred;
     }
     else if (syncType == "Periodic")
     {
--- a/src/data_sync_config.hpp
+++ b/src/data_sync_config.hpp
@@ -32,6 +32,7 @@ enum class SyncDirection
 enum class SyncType
 {
     Immediate,
+    Deferred,
     Periodic
 };
 
@@ -168,6 +169,8 @@ struct DataSyncConfig
         {
             case SyncType::Immediate:
                 return "Immediate";
+            case SyncType::Deferred:
+                return "Deferred";
             case SyncType::Periodic:
                 return "Periodic";
         }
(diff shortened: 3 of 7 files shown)


src/data_sync_config.hpp
```diff
@@ -253,6 +263,16 @@ struct DataSyncConfig
      */
     mutable std::unordered_set<fs::path> _syncInProgressPaths;
 
+    /**
```
RameshIyyar, line 266: Please push changes into relevant commits. I believe this commit is all about config parser, right?

src/manager.cpp
```diff
@@ -291,6 +291,28 @@ sdbusplus::async::task<> Manager::startSyncEvents()
                 setSyncEventsHealth(SyncEventsHealth::Critical);
```
RameshIyyar, line 291: Commit Message:

Add some example like PEL delete all use case.

> When a deferred event is received, record the latest event time
and schedule a deferred task if one is not already running. The
task waits until no new event arrives for the configured delay,
then runs one sync for the configured path.

When a data change event is received, record the latest event time
and schedule a deferred task if one is not already running. The
task waits until no new event arrives for the configured delay,
then runs one sync for the configured path instead of data change event path.

> If another event arrives while sync is running, the event time
is updated. After the current sync finishes, the task waits
again from that latest event time before running another sync.

If another event arrives while sync is deferred, the event time
is updated. The deferred time is always calculated from the remaining
deferral interval while the system sleeps for the deferral interval.
After the current sync finishes, the task waits
again from that latest event time before running another sync.

src/manager.hpp
```diff
@@ -279,6 +279,29 @@ class Manager
...
+     */
+    void deferSync(const config::DataSyncConfig& dataSyncCfg);
+
+    /**
+     * @brief Wait for the deferred interval and run sync.
+     *
+     * @param[in] dataSyncCfg - The data sync config to sync
+     */
+    sdbusplus::async::task<>
+        runDeferredSync(const config::DataSyncConfig& dataSyncCfg);
```
RameshIyyar: > runDeferredSync

syncDeferredData?

src/manager.cpp
```diff
@@ -819,6 +841,113 @@ sdbusplus::async::task<>
...
+    dataSyncCfg._deferredSyncScheduled = false;
+    co_return;
+}
+
+sdbusplus::async::task<>
+    // NOLINTNEXTLINE
+    Manager::monitorDeferredDataToSync(
+        const config::DataSyncConfig& dataSyncCfg)
+{
+    bool exception{false};
```
RameshIyyar, line 907: is it required?

src/manager.cpp
```diff
@@ -819,6 +841,113 @@ sdbusplus::async::task<>
...
+
+    dataSyncCfg._deferredSyncScheduled = true;
+    _ctx.spawn(runDeferredSync(dataSyncCfg));
+}
+
+sdbusplus::async::task<>
+    Manager::runDeferredSync(const config::DataSyncConfig& dataSyncCfg)
+{
+    while (!_ctx.stop_requested() && !_syncBMCDataIface.disable_sync() &&
+           dataSyncCfg._deferredSyncIntervalInSec.has_value())
```
RameshIyyar, line 860: why is it required?

config/schema/example.json
```diff
@@ -55,6 +55,13 @@
                 "Mode": "DBus",
                 "NotifyServices": ["Service1", "Service2"]
             }
+        },
```
RameshIyyar, line 58: I think this commit should come first.

src/manager.cpp
manishtwari: agree, this is not really required here

src/manager.cpp
manishtwari: If any failure happens in the deferred sync watcher flow, then it use to create an error log after the catch block.

src/manager.cpp
manishtwari: my bad, i remembered it is required cuz the clang is complaining with
unchecked access to optional value [bugprone-unchecked-optional-access,-warnings-as-errors]

RameshIyyar (approved):

src/data_sync_config.cpp
```diff
@@ -75,6 +75,19 @@ DataSyncConfig::DataSyncConfig(const nlohmann::json& config,
         _periodicityInSec = std::nullopt;
     }
 
+    if (_syncType == SyncType::Deferred)
+    {
+        constexpr auto defDeferredSyncInterval = 1;
```
aryakpadman, line 80: Just a suggestion, may be we could make this as build time config , so that user can override if they want .

manishtwari: DeferredSyncInterval is already configurable in JSON, so users can override it there.
https://github.com/ibm-openbmc/phosphor-data-sync/pull/113/changes/76602e24502484cd7eeabeed365e484e6f28db49

src/manager.cpp
```diff
@@ -745,32 +767,38 @@ sdbusplus::async::task<>
...
+                           ? std::make_optional<std::unordered_set<fs::path>>(
+                                 dataSyncCfg._excludeList.value().first)
+                           : std::nullopt;
+
+    auto [it, inserted] = _activeWatchers.emplace(
+        dataSyncCfg._path,
+        std::make_unique<watch::inotify::DataWatcher>(
+            _ctx, IN_NONBLOCK | IN_CLOEXEC, eventMasksToWatch,
+            dataSyncCfg._path, excludeList, dataSyncCfg._includeList));
+    static_cast<void>(inserted);
```
spinler: was there an actual error for an unused var here?
If so, also using '_' in the structured binding as the name works to silence it I've found.

spinler (approved):

src/manager.cpp
manishtwari: I did not hit an error locally, but `inserted` is unused here. I can use `_` in the structured binding and remove the `static_cast`.

src/data_sync_config.cpp
aryakpadman: Yeah , I was referring configuring the default defer interval also. 

I feel it's fine. We could consider later if use case arise.

aryakpadman (approved):

PR #115, opened 2026-08-10
inotify: stop active watchers when sync is disabled

manishtwari:
When sync is disabled, active DataWatcher instances may remain blocked waiting for an inotify event in onDataChange(). This prevents the monitor coroutines from exiting immediately.

Add DataWatcher::stop() to remove the active inotify watch descriptors. Removing a watch generates an IN_IGNORED event, which makes the inotify file descriptor readable and wakes the pending fdio::next() operation.

IN_IGNORED is not treated as a data change, so no sync operation is triggered. Once onDataChange() returns, the monitor loop observes that sync is disabled and exits. The existing scope cleanup then removes the DataWatcher from the active watcher map.

During failover, the passive BMC becomes active without restarting the application. Any watcher created for the previous BMC role may therefore remain active. Stopping the existing watchers ensures that they are removed before sync events are started again for the new BMC role.

Call stop() for all active watchers when the DisableSync property is set.


aryakpadman (approved):

src/data_watcher.cpp
```diff
@@ -56,6 +56,17 @@ int DataWatcher::inotifyInit() const
     return fd;
 }
 
+void DataWatcher::stop()
```
aryakpadman, line 59: Commit message :  Could you please mention what happens when sync is re-enabled without this change too ?

riya461 (approved):

spinler (approved):

RameshIyyar (approved):

